PR #191, opened 2016-09-29
Add missing packages

pi0:
**UPDATED** Some packages are missing during `npm run dist`. (*when clean repo cloned*) Packages including : 
+ async-validator
+ gulp-cssmin
+ gulp-postcss
+ throttle-debounce

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "url": "https://github.com/elemefe/element/issues"
   },
   "devDependencies": {
+    "async-validator": "^1.6.6",
     "babel-cli": "^6.14.0",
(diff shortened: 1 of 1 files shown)


Leopoldthecoder:
Run `npm run dist` instead of `gulp dist`. The former command installs these packages automatically.

pi0:
@Leopoldthecoder but this forces using `cooking` ! This makes gulp working out-of-the-box :) What's your idea?

QingWei-Li:
Each package is independent. You should run `npm run bootstrap` install dependencies

PR #312, opened 2016-10-10
feat(translation): date/time picker translation possibility if using …

karol-f:
Hi, I would like to add possibility to translate date/time pickers using vue-i18n.

- I've added detection - if vue-i18n exist we will use it's translations. If it's not available use built-in ones (`$t` helper) 
- move hardcoded language strings to translation file (`i18n.js`)

Regards


mygnu:
Awesome

PR #932, opened 2016-11-09
docs improve:el-input events

sindu12jun:
The el-input doc did not show the event when input is blured


Leopoldthecoder:
Thanks for contributing. We noticed some other error because of this PR and fixed them altogether in #936 .

PR #1117, opened 2016-11-16
fr translation

waam:
Add french translation.

### DIFF
--- a/src/locale/lang/fr.js
+++ b/src/locale/lang/fr.js
@@ -2,7 +2,7 @@ export default {
   el: {
     datepicker: {
       now: 'Maintenant',
-      today: "Auj.",
+      today: 'Auj.',
       cancel: 'Annuler',
       clear: 'Effacer',
       confirm: 'OK',
--- a/src/locale/lang/fr.js
+++ b/src/locale/lang/fr.js
@@ -2,16 +2,16 @@ export default {
   el: {
     datepicker: {
       now: 'Maintenant',
-      today: "Aujourd'hui",
+      today: "Auj.",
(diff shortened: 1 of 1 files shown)


src/locale/lang/fr.js
```diff
@@ -0,0 +1,85 @@
+export default {
+  el: {
+    datepicker: {
+      now: 'Maintenant',
+      today: "Aujourd'hui",
```
Leopoldthecoder: First, thank you for contributing! Here's a thing: some of the texts here are nested in a limited space. Take "Today" for example:
image
I don't think there is enough space for "Aujourd'hui".
Does the French language have any abbreviation or shorthand for "Aujourd'hui" and other long texts you provided here?

waam: "Auj." (with the dot at the end) could be used. It's popular.
"Mnt" is one possible abbreviation for "Maintenant", but it sounds weird aside the "OK" button.

"Sélectionner" could be replaced by "Choisir".

Leopoldthecoder: I see. So could you please replace these words with their appropriate abbreviations? Thanks.

src/locale/lang/fr.js
```diff
@@ -0,0 +1,85 @@
...
+      cancel: 'Annuler',
+      error: 'Erreur'
+    },
+    upload: {
+      delete: 'Supprimer',
+      preview: 'Aperçu',
+      continue: 'Continuer'
+    },
+    table: {
+      emptyText: 'Aucune donné',
```
jaoued: Aucune donnée

PR #1122, opened 2016-11-16
Update fr.js

bioleyl:
Small Spelling correction


bioleyl:
I had not seen that the correction was already in the pipe

PR #1406, opened 2016-11-28
support column filter

njleonzhang:
add props `column-input-filter` for el-table, which make el-table support input filter for every column.
image

*  `column-input-filter` can be boolen or string 'custom', which means el-table will not handle the filter. 
* `row.property` of the column is used to filter if `column.filterMethod ` is not defined. 
*  If `column.filterMethod ` is defined, it is used for the filtering. 
* `column-filter-change ` event is emitted with the newest column filters when any column filter changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
 - [Home Page](http://element.eleme.io/)
 - [Docs](http://element.eleme.io/#/component)
 - [FAQ](./FAQ.md)
+- [Customize Theme](http://element.eleme.io/#/en-US/component/custom-theme)
 - Starter Kit
  - [element-starter](https://github.com/ElementUI/element-starter)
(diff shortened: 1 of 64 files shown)


njleonzhang:
@Leopoldthecoder 怎么没人管。。。

Leopoldthecoder:
感谢贡献，不过设计和交互上可能不会接受这个功能。

myzingy:
后台管理界面很期望这个功能，检索方便

njleonzhang:
@myzingy 貌似只能自己实现，自己维护了 😂

myzingy:
@njleonzhang 我怎么把你的push，弄到我的项目中，有空了写个大概的文档，非常感谢你的贡献

njleonzhang:
@myzingy 这是git的操作, 你去google一下吧。

PR #1436, opened 2016-11-28
Created tr-TR.js for Turkish Language Support

barbaroskurt:
Created tr-TR.js for Turkish Language Support


src/locale/lang/tr-TR.js
```diff
@@ -0,0 +1,89 @@
+'use strict';
+
+exports.__esModule = true;
+exports.default = {
```
QingWei-Li: Please use ES2015, such as `export default =` 😄

PR #1545, opened 2016-12-04
Included pt-br locale

vedovelli:


Leopoldthecoder:
Thanks for contributing!

Could you tell me what the difference is bewteen pt-br and [pt](https://github.com/ElemeFE/element/blob/dev/src/locale/lang/pt.js)?

vedovelli:
@Leopoldthecoder Hello! There are slight differences from pt (which is from Portugal) and pt-br (which is from Brazil). Thanks!

emtudo:
thanks @vedovelli abs!

kaibady:
Why is this lowercase

PR #1781, opened 2016-12-15
Add Indonesian translation

rizqyhi:
Hi,
just translating this awesome project to Indonesian.
Thank you

### DIFF
--- a/src/locale/lang/id.js
+++ b/src/locale/lang/id.js
@@ -1,0 +1,86 @@
+export default {
+  el: {
+    datepicker: {
+      now: 'Sekarang',
+      today: 'Hari ini',
+      cancel: 'Batal',
+      clear: 'Kosongkan',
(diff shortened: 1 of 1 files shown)


src/locale/lang/id.js
```diff
@@ -0,0 +1,86 @@
+export default {
+  el: {
+    datepicker: {
+      now: 'Sekarang',
+      today: 'Hari ini',
```
Leopoldthecoder: Thanks for contributing!

There may not be enough space for the eight-lettered`Hari ini`, do you have any shorthand for it in Indonesian?

rizqyhi: I'm afraid there's no equal word other than `hari ini`. But may be we can use shorted word of `sekarang` which is `Skrng`. It's not usual but I guess people will got it.

PR #1873, opened 2016-12-20
Update input.css

Jokcy:
input border radius should use root variable

### DIFF
--- a/packages/theme-default/src/input.css
+++ b/packages/theme-default/src/input.css
@@ -10,7 +10,7 @@
       appearance: none;
       background-color: var(--color-white);
       background-image: none;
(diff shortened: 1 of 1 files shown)


baiyaaaaa:
应该是 `--input-border-radius`  才对

packages/theme-default/src/input.css
```diff
@@ -10,7 +10,7 @@
       appearance: none;
       background-color: var(--color-white);
       background-image: none;
-      border-radius: 4px;
+      border-radius: var(--border-radius-small);
```
baiyaaaaa: 应该是 --input-border-radius 才对?

PR #1967, opened 2016-12-23
Col: Fix issue where xs is overridden by general

azurelogic:
The programmatic generation of xs and general column widths together causes .el-col-24 (and other higher column numbers) to override basically all (lower column number) .el-col-xs-XX styles. Separating them fixes this issue.

### DIFF
--- a/packages/theme-default/src/col.css
+++ b/packages/theme-default/src/col.css
@@ -7,20 +7,33 @@
 }
 
 @for $i from 1 to 24 {
-  .el-col-$i,
+  .el-col-$i {
+    width: calc(1 / 24 * $(i) * 100)%;
+  }
+  .el-col-offset-$i {
+    margin-left: calc(1 / 24 * $(i) * 100)%;
+  }
+  .el-col-pull-$i {
+    position: relative;
+    right: calc(1 / 24 * $(i) * 100)%;
+  }
+  .el-col-push-$i {
+    position: relative;
+    left: calc(1 / 24 * $(i) * 100)%;
+  }
+}
+
+@for $i from 1 to 24 {
   .el-col-xs-$i {
     width: calc(1 / 24 * $(i) * 100)%;
   }
-  .el-col-offset-$i,
   .el-col-xs-offset-$i {
     margin-left: calc(1 / 24 * $(i) * 100)%;
   }
-  .el-col-pull-$i,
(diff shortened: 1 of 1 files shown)


packages/theme-default/src/col.css
```diff
@@ -7,20 +7,33 @@
...
+    position: relative;
+    right: calc(1 / 24 * $(i) * 100)%;
+  }
+  .el-col-push-$i {
+    position: relative;
+    left: calc(1 / 24 * $(i) * 100)%;
+  }
+}
+
+@for $i from 1 to 24 {
```
baiyaaaaa: need to be wrap in media query

@media (max-width: 768px) {
  @for $i from 1 to 24 {
  }
}

baiyaaaaa: need to be wrap in media query

@media (max-width: 768px) {
@for $i from 1 to 24 {
}
}

baiyaaaaa: need to be wrap in media query

@media (max-width: 768px) {
@for $i from 1 to 24 {
}
}

azurelogic: @baiyaaaaa: This would cause breaking changes from the expected responsive behavior. In most responsive grid frameworks, smaller viewports are defined first and are used as the fallback for larger viewports, which appeared to be the original intent here. Note that none of the other breakpoints have max-width set. If you're concerned about the .el-col-xs-XX classes overriding the general .el-col-XX classes, then a more appropriate solution would be to remove one of them (probably the xs classes) and use them when the xs prop is set on a column. See how Bootstrap does it here: old way = https://github.com/twbs/bootstrap/blob/v4.0.0-alpha.5/dist/css/bootstrap.css and new way (appears to be bleeding edge and not in the v4 docs yet) = https://github.com/twbs/bootstrap/blob/v4-dev/dist/css/bootstrap.css. Please reconsider. Thanks!

baiyaaaaa: @azurelogic Thanks!Your suggestion is very valuable to me. Indeed, i just concerned about the .el-col-xs-XX classes overriding the general .el-col-XX classes. I think i should use the way as you've said. it will be fix in next patch version. However, I still have a question, what's the different between .el-col-XX and .el-col-xs-XX classes?

PR #2145, opened 2017-01-02
InputNumber, DatePicker: Clearify the logic of increase and decrease. Fix date-picker.spec.js bugs.

liril-net:
1. Add `max` and `min` to `el-input`.
2. Replace `acAdd, accSub` with `_increase, _decrease`.
3. Add `precision`.
4. Change default `min` to `-Infinity`.


liril-net:
Fix bugs in `date-picker.spec.js` to pass the check.

PR #2173, opened 2017-01-03
Input: Add propagateEvent to emit event from children to parent.

liril-net:
1. remove something like `handleClick`, which only emit event to parent.
2. add propagateEvent to emit event to parent.


QingWei-Li:
Maybe we can use `.native` ?

```html
<el-input @keydown.native="handleEvent" />
```

liril-net:
Uh huh, u're right, i will close this pr.

PR #2191, opened 2017-01-04
Locale: customize i18n handler, close #2129

### LINKED ISSUE #2129: Conflict between vuei18n and vuex-i18n in $t function
Element-UI now supports **vuei18n** and is checking if the function `$t` is used in `/element-ui/src/locale.index.js`
```
export const t = function(path, options) {
  const vuei18n = Object.getPrototypeOf(this || Vue).$t;
  if (typeof vuei18n === 'function') {
    if (!merged) {
      merged = true;
      Vue.locale(
        Vue.config.lang,
        deepmerge(lang, Vue.locale(Vue.config.lang) || {}, { clone: true })
      );
    }
    return vuei18n.apply(this, [path, options]);
  }
  const array = path.split('.');
  let current = lang;

  for (var i = 0, j = array.length; i < j; i++) {
    var property = array[i];
    var value = current[property];
    if (i === j - 1) return format(value, options);
    if (!value) return '';
    current = value;
  }
  return '';
};
```
I'm not using this package but I'm using `vuex-i18n` which also implements the same function `$t` thus this condition `if (typeof vuei18n === 'function')` is true and code inside fails because  it's checking the wrong library.

This causes the error: `TypeError: _vue2.default.locale is not a function` 

I think supporting `vuei18n` should be optionnal and not there by default

### ElementUI version
1.0.9

### Vue version
2.1.6

QingWei-Li:

* [x] Make sure you follow Element's [Contributing Guide](https://github.com/ElemeFE/element/blob/master/.github/CONTRIBUTING.md).

### DIFF
--- a/examples/docs/zh-CN/i18n.md
+++ b/examples/docs/zh-CN/i18n.md
@@ -68,6 +68,22 @@ Vue.locale('zh-cn', zhLocale)
 Vue.locale('en', enLocale)
 ```
 
+## 兼容其他 i18n 插件
+如果不使用 `vue-i18n`，而是用其他的 i18n 插件，Element 将无法兼容，但是可以自定义 Element 的 i18n 的处理方法。
+
+```javascript
+import Vue from 'vue'
+import Element from 'element-ui'
+import enLocale from 'element-ui/lib/locale/lang/en'
+import zhLocale from 'element-ui/lib/locale/lang/zh-CN'
+
+Vue.use(Element, {
+  i18n: function (path, options) {
+    // ...
+  }
+})
+```
+
 目前 Element 内置了以下语言：
 <ul class="language-list">
   <li>简体中文（zh-CN）</li>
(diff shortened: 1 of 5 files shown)


examples/docs/zh-CN/i18n.md
```diff
@@ -68,6 +68,22 @@ Vue.locale('zh-cn', zhLocale)
 Vue.locale('en', enLocale)
 ```
 
+## 兼容其他 i18n 插件
```
Leopoldthecoder: Compatibility with other i18n plugins

examples/docs/zh-CN/i18n.md
```diff
@@ -68,6 +68,22 @@ Vue.locale('zh-cn', zhLocale)
...
+如果不使用 `vue-i18n`，而是用其他的 i18n 插件，Element 将无法兼容，但是可以自定义 Element 的 i18n 的处理方法。
```
Leopoldthecoder: Element may not be compatible with i18n plugins other than `vue-i18n`, but you can customize how Element processes i18n.

PR #2207, opened 2017-01-05
table: wrap col in colgroup to make column width work in safari

reverland:
I dont know whether it is about #2198 

but I find some of my tables width is totally messed in safari.

I try to wrap `<col>` in `<colgroup>` and it works in safari now.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/col


Kingwl:
dalao holly high

Issue #2332, opened 2017-01-10
el-select with multiple and filterable options combined


zeratulmdq:
### ElementUI version
1.1.3 (though I guess all versions)

### OS/Browers version
Linux/Firefox 50.1.0 (though I guess all versions)

### Vue version
2.1.0 (though I guess all versions)

### Reproduction Link
https://jsfiddle.net/8o1eqbd6/

### Steps to reproduce
Use an el-select element with *multiple* and *filterable* options combined.

### What is Expected?
If you start typing (let's say 'T') and then click on *Two*, that option gets selected, the input should be focused again and the text erased.

### What is actually happening?
After you click on *Two* the input looses focus, you can't use backspace to erase, the text is still there and if you click outside the component or inside the input, the list disappears so you have to click again. It's annoying when you want to select multiple options (tags for a post in my case)

Leopoldthecoder:
Fixed in https://github.com/ElemeFE/element/pull/2540/commits/8fc55e30e6ef2611e5a54df6488db4856e7943c1

Now after selecting an option, the input regains focus. But the text is not erased because in some cases you may need to select more than one option for the same query.

zeratulmdq:
Thanks! Not a fan of the text part but it's ok. Great work BTW.

Leopoldthecoder:
En Taro Tassadar!

anotheri:
@Leopoldthecoder 
could that behaviour be optional to erase the query text?
or how can i force it to erase the query text?

allfreelancers:
@anotheri how did you solve this problem?

anotheri:
@allfreelancers 

your html (or jade, like in my case) template should contain `ref` attribute as well as use listener on `@change` event, like this:
```jade
el-select(
  ref="participants",
  v-model="form.participants",
  multiple,
  filterable,
  :default-first-option='true',
  @change='onChangeParticipants'
)
  el-option(
    v-for="person in people",
    :key="person.id",
    :value="person.id",
    :label="person.name"
  )
```

and you need to define that listener function like this in your js code:
```js
onChangeParticipants ( e, data ) {
  this.$refs.participants.query = '';
},
```

allfreelancers:
@anotheri Thank you so much!

Leopoldthecoder:
In version 2.0 we'll have a new prop `reserve-keyword`. When set to false, the query text will be cleared after selecting an option.

zeratulmdq:
A W E S O M E

[...]

PR #2474, opened 2017-01-17
Select: restore focus and erase query

gil0mendes:
After an option selection/creation the focus must be restored and the query erased. This problem was repeated on the issue #2332.

----

This works, but I tried run the tests and I get this message. How can I fix this?

```text
TypeError: undefined is not an object (evaluating 'this.$refs.input.focus') (webpack:///packages/select/src/select.vue?79b8:9:0 <- index.js:9696)
```

### DIFF
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -505,10 +505,13 @@
             this.value.push(option.value);
           }
           if (option.created) {
-            this.query = '';
             this.inputLength = 20;
-            this.$refs.input.focus();
           }
+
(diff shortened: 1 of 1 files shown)


Leopoldthecoder:
Will look into this soon, though I'm not quite sure erasing query is necessary (totally agree that focus should be restored).

Leopoldthecoder:
https://github.com/ElemeFE/element/pull/2540/commits/8fc55e30e6ef2611e5a54df6488db4856e7943c1

Issue #2561, opened 2017-01-21
el-popover leaking memory


badpunman:
```
<el-popover ref="popover1" title="Title" trigger="click">
   <div slot="reference">Click this</div>
</el-popover>
```

Basically, just create this, and move out of the component with vue-router(in my test case) to remove it, and there is some memory leakage with detached dom-elements and apparently whole previous vuenode tree still in memory. The memory usage will especially start to add up on IE and Firefox with their implementation of the garbage collection. 

### ElementUI version
1.1.5

### OS/Browers version
all browsers

### Vue version
2.1.10

badpunman:
#2570 attempt at fix. Removing document click handlers that seemed to pile up.

PR #2570, opened 2017-01-22
remove popover's document click handler, related to #2561

badpunman:
This doesn't actually seem to fix the problem completely, but it should be a start.
There are still some VueComponents of the popovers in memory.


badpunman:
My memory leak seemed to attributed to either vuerouter or my build in general. Ignore the previous whinings of popover leaking, besides the document click handler.

superchangme:
i found it still has memory leak i use it with elTree, the huge memory cant release after popover closeed

PR #2608, opened 2017-01-25
Close icon design adjustment for removable tags

e0:
I noticed that when hovering over the close icon in a removable tag, the `x` is a little too close to the top right of the circle, to the point where it is touching the edge. This change makes it appear more centered in the circle. It still doesn't look perfectly centered but at least to me it is more visually appealing. 

Before: 
tag_remove_before

After:
tag_remove_after


Leopoldthecoder:
Thanks, but changing font size does not seem to be the best way to fix this because in some browsers, font sizes smaller than 12px actually make no difference.

PR #2629, opened 2017-01-30
Update component.vue

antocorso:
I think that in the title I should put html so for example I can put an image in the title!! (I need to put a flag image for example for translation purpose)


antocorso:
ok, thanks

Leopoldthecoder:
#2657

PR #2870, opened 2017-02-16
Form submit causes page reload

antarasi:
**Problem case 1:**
We have `<el-button native-type="submit" .... >` inside `el-form`. Then after hiting enter while editing some text input, browser tries to submit the form hence the page is reloaded becuse we do not have submit event handler attached to el-form.

**Problem case 2:**
We have `<el-button native-type="button" .... >` inside `el-form` hence there is no submit input/button. Then hiting enter while editing some input does nothing and we must use button click to submit form (in JavaScript).

**To sum up:** `el-form` does not have @submit event support. Solution in this pull request is not elegant but, solves the problem with regular form submit and allows to use submit input inside form.

Better solution would be to add @submit and @submit.prevent handler to use as parameter.

* [ ] Make sure you follow Element's contributing guide ([中文](https://github.com/ElemeFE/element/blob/master/.github/CONTRIBUTING.zh-CN.md) | [English](https://github.com/ElemeFE/element/blob/master/.github/CONTRIBUTING.en-US.md)).

### DIFF
--- a/packages/form/src/form.vue
+++ b/packages/form/src/form.vue
@@ -2,7 +2,7 @@
   <form class="el-form" :class="[
     labelPosition ? 'el-form--label-' + labelPosition : '',
     { 'el-form--inline': inline }
-  ]">
+  ]" onsubmit="return false;">
     <slot></slot>
   </form>
 </template>


Leopoldthecoder:
I guess adding `@submit.native.prevent` to the `<el-form>` tag would do the job. See the first question in our FAQ for details.

antarasi:
@Leopoldthecoder Yes, it's working, thank you! By the way I think listing FAQ topics directly on main readme would lower the number of issues posted.

kkinder:
Am I missing something? It seems to me that using `@submit.native.prevent` *does not* seem to work:

https://jsfiddle.net/19bwhak8/2/

Leopoldthecoder:
@kkinder Pressing enter submits the form only when the form has one input element: https://jsfiddle.net/19bwhak8/3/ . This is what native HTML does.

kkinder:
@Leopoldthecoder ah, it's because <el-button type="submit"> isn't sending the submit. If you add a hidden submit button, it works.

PR #3080, opened 2017-02-26
Slider:Prevent text selection on mouse drag

PanJiaChen:
修复Slider drag时会选中周围text

### DIFF
--- a/packages/slider/src/button.vue
+++ b/packages/slider/src/button.vue
@@ -96,6 +96,7 @@
 
       onButtonDown(event) {
         if (this.disabled) return;
+        event.preventDefault();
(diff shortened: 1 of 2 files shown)


Leopoldthecoder:
感谢贡献。测试没通过，需要在[这里](https://github.com/ElemeFE/element/blob/dev/test/unit/specs/slider.spec.js#L80)和[这里](https://github.com/ElemeFE/element/blob/dev/test/unit/specs/slider.spec.js#L104)分别加上 `preventDefault() {}`

PanJiaChen:
帮忙看看 我这样加test对不对

PR #3345, opened 2017-03-08
Loading: keep loading text sync with element-loading-text attribute

imyzf:
Get value from `element-loading-text` attribute on directive update hook, then we can update loading text dynamically.

Fix issue #2272.

### DIFF
--- a/packages/loading/src/directive.js
+++ b/packages/loading/src/directive.js
@@ -91,6 +91,7 @@ exports.install = Vue => {
 
     update: function(el, binding) {
       if (binding.oldValue !== binding.value) {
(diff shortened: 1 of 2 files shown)


packages/loading/src/directive.js
```diff
@@ -91,6 +91,7 @@ exports.install = Vue => {
 
     update: function(el, binding) {
       if (binding.oldValue !== binding.value) {
+        el.instance.setText(el.getAttribute('element-loading-text'));
```
Leopoldthecoder: Thanks for contributing. This line should go outside the `if` block because everything inside this block will not execute until the binding value changes.

Issue #3121, opened 2017-02-27
popover内嵌select组件,点击popover空白区域select下拉框不会消失


Yuliang-Lee:
popover内嵌select组件,点击popover空白区域select下拉框不会消失

### Reproduction Link
http://jsfiddle.net/2p1gom7n/

Tronsfey:
Came across the same issue :)

amnsss:
Came across the same issue :)

amnsss:
这个有啥快速的hack办法吗

jack-liyou:
Came across the same issue :)

Leopoldthecoder:
Fixed in #3451

PR #3451, opened 2017-03-12
Clickoutside: prevent closing on drag click

nicoeg:
When eg. having to choose a color at the edge of the Colorpicker component, it is easy to let your mouse get out of the components box. This will make the Colorpicker disappear which can be frustrating for the user.

mar -12-2017 21-45-54

I've added a mousedown event in the Clickoutside directive to check whether the mouse' starting point is inside the "popperElm" and prevent the bindings method to be called if it is.

This might add some unwanted results in other components using the clickoutside but shouldn't really be an issue. 
If it is though, I propose adding a modifier to the directive to signal that drag clicking shouldn't call the bindings method. `<element v-clickoutside.allowdrag="hide">`

Note: The test utility method 'triggerClick's  description is translated with Google Translate. So you might want to correct me on this one 😄 

Let me know what you think.

\- Thanks for this awesome project

### DIFF
--- a/src/utils/clickoutside.js
+++ b/src/utils/clickoutside.js
@@ -9,7 +9,7 @@ let startClick;
 !Vue.prototype.$isServer && on(document, 'mousedown', e => (startClick = e));
 
 !Vue.prototype.$isServer && on(document, 'mouseup', e => {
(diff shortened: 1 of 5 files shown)


Leopoldthecoder:
Thanks! This also fixes #3121.

Issue #3473, opened 2017-03-13
[Bug]: Set slider step after a few moments doesn't update precision


gabrielboliveira:
### ElementUI version
1.2.4

### OS/Browsers version
Chrome 56

### Vue version
2.2.3

### Reproduction Link
http://jsfiddle.net/1xxkymte/

### Steps to reproduce
Change slider step after a few moments it's mounted, it will not update the precision and it will not work as expected

### What is Expected?
Change slider step should update precision. Example: updating step from 10 to 0.01 after 100ms it's mounted, should update steps, but actually it stays on 1

### What is actually happening?
Precision is always 1.

PR #3475, opened 2017-03-13
Slider: fix step not updating precision

gabrielboliveira:
* [X] Make sure you follow Element's contributing guide

Fixing #3473. When `step` was updated on parent component, Slider didn't update the precision, leading to some errors.

### DIFF
--- a/packages/slider/src/main.vue
+++ b/packages/slider/src/main.vue
@@ -99,7 +99,6 @@
         firstValue: null,
         secondValue: null,
         oldValue: null,
-        precision: 0,
         dragging: false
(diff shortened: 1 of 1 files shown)


Leopoldthecoder:
Thanks!

This fix actually reminds me that it would be better if `this.precision` is a computed property, so that it can be updated whenever `this.min`, `this.max` or `this.step` changes.

So could you update this fix?

gabrielboliveira:
@Leopoldthecoder great idea! I've changed `precision` to be a computed prop, please take a look :)

PR #3597, opened 2017-03-19
Icon: Change fonts URL to absolute

fntneves:
The current font URL has unexpected results.

Example: If browser is in URL http://app.com/users/create, fonts will be loaded from http://app.com/users/create/fonts...., which is likely to return a 404.

This PR proposes adding a slash to the beginning of each font URL to ensure always the same behavior.

Example:
http://app.com/users/create -> http://app.com/fonts....
http://app.com/ -> http://app.com/fonts....

### DIFF
--- a/packages/theme-default/src/icon.css
+++ b/packages/theme-default/src/icon.css
@@ -1,7 +1,7 @@
 @font-face {
     font-family: 'element-icons';
-    src: url('fonts/element-icons.woff?t=1472440741') format('woff'), /* chrome, firefox */
-      url('fonts/element-icons.ttf?t=1472440741') format('truetype'); /* chrome, firefox, opera, Safari, Android, iOS 4.2+*/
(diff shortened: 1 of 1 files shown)


Leopoldthecoder:
Could tell us how you're importing Element? Because normally this is not an issue, and this fix actually breaks our test.

skrichten:
I have run into problem like this.  If the user reloads the page when they are at a URL that is deeper than the base URL, the font will not load.  For example a site with an entry URL like this  ```/site/home``` and the user reloads the page while on a URL like this ```/site/products/1234``` ... the font doesn't load.

hitoridewanai:
+1 on this. Also breaks build systems where css is in different directory than fonts (phoenix + brunch).

skrichten:
In my case, I was able to fix the problem by adding `` publicPath: "/" `` to my webpack output options.  I think it was this stack overflow post the helped me figure it out...

https://stackoverflow.com/questions/34133808/webpack-ots-parsing-error-loading-fonts?rq=1

PR #3703, opened 2017-03-22
Theme: Add missing zero-offset classes to responsive column stylesheet

JanCVanB:
This fixes a bug where a div with both
`.col-sm-offset-1` and `.col-md-offset-0` classes
will have the `small-offset-1` value for `margin-left`
in a medium-width window.

### DIFF
--- a/packages/theme-default/src/col.css
+++ b/packages/theme-default/src/col.css
@@ -8,6 +8,9 @@
 .el-col-0 {
   display: none;
 }
+.el-col-offset-0 {
+  margin-left: 0;
+}
 
 @for $i from 1 to 24 {
   .el-col-$i {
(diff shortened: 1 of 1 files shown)


packages/theme-default/src/col.css
```diff
@@ -8,6 +8,9 @@
 .el-col-0 {
   display: none;
 }
+.el-col-offset-0 {
+  margin-left: 0;
+}
 
 @for $i from 1 to 24 {
```
JanCVanB: Should we just change these for loops to `@for $i from 0 to 24`? I don't know if the `pull-0` and `push-0` need this "missing zero-classes" fix as well.

baiyaaaaa: i think it's a good idea `change these for loops to @for $i from 0 to 24`

JanCVanB: cool, thanks! (it was actually #4053)

PR #3781, opened 2017-03-25
Input: Add Keyup.Enter event

hughsaffar:
It is a common practice for input elements to have keyup.enter event available. For many instances it is handy to listen for this event as many user tends to use enter key for an action.


hughsaffar:
My bad, didn't know about the keyup.native.enter. But you are right, if we add one event we have to have them all. I am going to work on a generic method could handle all type of keyup events.

PR #3940, opened 2017-03-31
Ensure inner el-upload uses name attribute

mega667:
Without the name or id property the form field cannot be submitted as a regular non-xhr ajax request. I've opted for adding just the name attribute.

The only issue I see with this is that the `handleChange` event will not be called again if a user selects the same file (since path / name is unchanged).


Leopoldthecoder:
@baiyaaaaa Please take a look.

PR #4230, opened 2017-04-15
Quickstart.md: Fixed missing component imports in the Quickstart exam…

lfan06:
Fixed the missing component imports in the Quick Start example page (http://element.eleme.io/#/en-US/component/quickstart), without them the client page gets js errors that prevent vue components from loading.


Leopoldthecoder:
Thanks. Could you also fix the Chinese Quickstart.md?

lfan06:
Also fixed the Chinese Quickstart.md.

PR #4282, opened 2017-04-18
Upload: fix uploader can't disable

xingoxu:

The upload component attched the clickHandler to the parent el so the disabled button can't work properly. This commit fixed this by add clickHanlder on slot.

A more elegant way to disable the uploader is adding disable prop to the uploader, but that may be an another feature request.

### DIFF
--- a/packages/upload/src/upload.vue
+++ b/packages/upload/src/upload.vue
@@ -135,12 +135,19 @@ export default {
     const data = {
       class: {
         'el-upload': true
-      },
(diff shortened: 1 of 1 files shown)


CYvonne:
最近我也是遇到了这个问题，等待下一版本...

baiyaaaaa:
HI thanks for advise! I add the disabled feature in #4369. I think add disable prop to the uploader is a better choice.

Issue #4153, opened 2017-04-12
[Feature Request] Add text to the carousel indicator buttons


paul-blundell:
### Existing Component
Yes

### Component Name
Carousel

### Description
The small lines under the carousel are buttons, it would be very useful if we could add text to those buttons, a title of the carousel item such as Offers, etc.
See this example:

![screen shot 2017-04-12 at 09 09 34](https://cloud.githubusercontent.com/assets/1451631/24947952/9900ee50-1f60-11e7-9984-93eb4201e256.png)

Leopoldthecoder:
We'd like to keep Carousel simple, so no plan to add this.

paul-blundell:
Unless I'm mistaken, this won't add any extra complexity and would be a completely optional thing.
You can just add a prop to the carousel-item for label and then modify main.vue so when it displays the indicators it adds the value to the button:

`<button class="el-carousel__button">{{ item.label }}</button>`

Or, can you let me know how I can create a carousel similar to the following. Element has so far been fantastic and would prefer to stick with it for this too.

![screenshot 2017-04-17 at 16 42 14](https://cloud.githubusercontent.com/assets/1451631/25094033/1d260dac-238d-11e7-837b-daebd7de1baf.png)

chixiaoxiao:
我也是有碰到这个问题，想要自己设置下面按钮的样式显示效果之类的，横线暂时不太适用，不能和其他的一样自己需要的话，自己编辑一段替换插入嘛？

paul-blundell:
I have submitted a pull request for this, https://github.com/ElemeFE/element/pull/4317

Hopefully the developers will consider it. It would be a completely optional feature.

PR #4317, opened 2017-04-19
[Carousel] Add an optional label to indicator buttons

paul-blundell:

* [x] Make sure you follow Element's contributing guide

Adds an optional label to the indicator buttons on the carousel. Can be used only if required like so:

```
<n-carousel-item label="Slide 1">
  <h3>My Slide</h3>
</n-carousel-item>
```

### DIFF
--- a/packages/carousel/src/item.vue
+++ b/packages/carousel/src/item.vue
@@ -30,7 +30,10 @@
 
     props: {
       name: String,
-      label: String
+      label: {
+        type: [String, Number],
+        default: ''
+      }
     },
 
     data() {
--- a/packages/carousel/src/item.vue
+++ b/packages/carousel/src/item.vue
@@ -29,7 +29,8 @@
     name: 'ElCarouselItem',
 
     props: {
-      name: String
+      name: String,
+      label: String
     },
 
     data() {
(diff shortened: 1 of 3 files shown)


paul-blundell:
That was a fair point, I didn't think about the default theme.

I have added a new prop (defaults to false) for indicator-labels on the carousel itself. If true then it applies another class to the button and adds the label text from carousel-item.

Please let me know your thoughts.

Leopoldthecoder:
image

I tried the new commits and got the above. It doesn't seem to be desired. 

To get what you proposed in #4153, I think more CSS needs to be added. For example, if `indicator-labels` is true, another class should be applied to `<ul>` to deal with its positioning and width. Besides, since the indicators can be both inside and outside the slides (governed by `indicator-position`), it makes sense that labelled indicators should look nice in both scenarios.

What do you think? Could you update this PR?

paul-blundell:
Updated to fix that issue. Added a class to the `<ul>` instead like you suggested.
There is still some overlap on smaller screens when you have a lot of carousel items, but maybe this should be down to the user to style as required or limit the number of items?

Leopoldthecoder:
I updated this. Removed `indicatorLabels` because we can check if items have `label`s inside main.vue.

PR #4634, opened 2017-05-03
Form: Add label slot

liril-net:

### DIFF
--- a/packages/form/src/form-item.vue
+++ b/packages/form/src/form-item.vue
@@ -5,7 +5,7 @@
     'is-required': isRequired || required
   }">
     <label :for="prop" class="el-form-item__label" v-bind:style="labelStyle" v-if="label">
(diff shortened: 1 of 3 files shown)


packages/form/src/form-item.vue
```diff
@@ -5,7 +5,7 @@
     'is-required': isRequired || required
   }">
     <label :for="prop" class="el-form-item__label" v-bind:style="labelStyle" v-if="label">
```
breeswish: What if there are only label slots but no label props?

daggerjames: the slot won't be appended. and I think it is unnecessary to put label="" for this usage

WLyKan: It seems that this is the easiest solution, `<label>`is necessary in label.

PR #4780, opened 2017-05-10
[ElUpload] Fix close icon postion and icons should be hidden when in upload proc…

lifeishell:
1. When use file upload, the closed icon shows in a wrong position. should move into the label element.
2. When progress bar shows, the icons has not been hidden.


baiyaaaaa:
please rebase your code

PR #4839, opened 2017-05-12
dialog function parameter => done(false)

patriciussanctus:
For using `done(false)` as well as `vue-router.beforeEach` similarly function `next(false)`.
There are cases when it is necessary.

### DIFF
--- a/packages/dialog/src/component.vue
+++ b/packages/dialog/src/component.vue
@@ -122,9 +122,11 @@
           this.hide();
         }
       },
-      hide() {
-        this.$emit('update:visible', false);
-        this.$emit('visible-change', false);
+      hide(cancel) {
+        if (cancel !== false) {
+          this.$emit('update:visible', false);
+          this.$emit('visible-change', false);
+        }
       },
       updatePopper() {
         this.broadcast('ElSelectDropdown', 'updatePopper');


Leopoldthecoder:
Could you give an example where this is necessary?

patriciussanctus:
i use this function at the same time for routerBeforeEach and for modalBeforeClose.

```
leaveConfirm ({ commit, state }, next) {
      return new Promise((resolve, reject) => {
        if (state.routeLeaveConfirm) {
          Vue.prototype.$confirm('Unsaved changed will be lost. Leave?', 'Warning', {
            confirmButtonText: 'Ok',
            cancelButtonText: 'Cancel',
            type: 'warning'
          }).then(() => {
            resolve()
            commit('UPD_ROUTELEAVE', false)
            next()
          }).catch(() => {
            // reject()
            commit('UPD_ROUTELEAVE', false)
            next(false)
          })
        } else {
          resolve()
          next()
        }
      })
    }
```

Leopoldthecoder:
Calling `next(false)` seems to have the same effect as simply not calling `next` at all?

patriciussanctus:
yea, but if only ```next(false)``` was absent, i would wallow in conditions...

Issue #4644, opened 2017-05-04
[Bug Report] attribute: before-close and done (dosen't) work dialog box before-close attribute does not work, done () does not work


fairyier:
### Element UI version
1.3.0-beta.3

### OS/Browsers version
win10/google

### Vue version
2.2.6

### Reproduction Link
[https://jsfiddle.net/api/post/library/pure/](https://jsfiddle.net/api/post/library/pure/)

### Steps to reproduce
点击打开dialog---打开对话框----点击确认或者取消按钮，都不会触发:before-close="handleClose"  ，点对话框右上角的关闭可以触发:before-close="handleClose"调用comfirm弹窗，但是点confirm的确认按钮，只关闭confirm弹窗本身，dialog对话框还在，done（）没起作用

### What is Expected?
点击确认或者取消按钮可以触发:before-close="handleClose"    ，点击confirm的确认按钮两个弹窗 都要关闭

### What is actually happening?
点击确认或者取消按钮没触发:before-close="handleClose"    ，点击confirm的确认按钮，只关闭confirm弹窗本身

Leopoldthecoder:
You have total control over the confirm and cancel buttons, so you can do whatever you want in their click handler.

Element 1.3.0 is released, and the `done` method can now close Dialog correctly.

Issue #4492, opened 2017-04-27
[Bug Report] this.$confirm: cancel callback doesn't not triggered when closed by Esc or click on the mask layer


Riant:
### Element UI version
1.2.9

### OS/Browsers version
Windows / Chrome56

### Vue version
2.3.0-beta.1

### Reproduction Link
[https://jsfiddle.net/pj71jkyw/142/](https://jsfiddle.net/pj71jkyw/142/)

### Steps to reproduce
1. click the open confirm button to open the confirm box
2. click on the keyboard on the Esc or pop-up box mask layer

### What is Expected?
Trigger cancel callback like click on the close button and cancel button

### What is actually happening?
No cancel callback is triggered, and no other options can be configured for this action callback

Leopoldthecoder:
Fixed [here](https://github.com/ElemeFE/element/pull/4851/files#diff-27b250184267261d8e837735fd53f227R132)

PR #4856, opened 2017-05-14
[Popup] Fix popup-manager to call beforeClose() when closeOnClickModal

JanCVanB:
This is to fix a bug I've experienced since 1.3.0 where I can't close dialogs by clicking on the mask/background.

I don't know if this is the right approach, I just copied it from #4819 - the fix to call beforeClose() when closeOnPressEscape.

This might help the following issues: #4458, #4492, #4642, #4644, #4699, #4700, #4701, #4784


Leopoldthecoder:
> I can't close dialogs by clicking on the mask/background

Could you write a demo?

PR #4874, opened 2017-05-15
Select: Blur non-filterable select to avoid keyboard on mobile like i…

karol-f:
Select: Blur non-filterable select to avoid keyboard on mobile like iOS or Firefox Mobile.

On mobile devices (iOS e.g. iPhone, Firefox Mobile), even if SELECT is not filterable, keyboard bar activates. This commit resolves this issue.

### DIFF
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -421,6 +421,10 @@
 
       handleFocus() {
         this.visible = true;
+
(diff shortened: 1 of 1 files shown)


Leopoldthecoder:
Element is not designed for mobile devices. So it's better to avoid pieces of code that are specifically for mobile devices. Actually you can deal with this keyboard problem in the `visible-change` event handler.

PR #5113, opened 2017-05-29
Dialog:  componentize it to provide common methods like MessageBox

Xiaoshan-Nian:
We need to dynamically make an instance of Dialog, but we don't want to bind the dialog with a tag in the template. So we need a dialog component like Messagebox, a dialog that can be open by $dialog/$modal like $msgbox, especially when a new dialog is generated by click some elements in an opened dialog. There can be a queue of dialogs. 

Although Messagebox to some extend meets this requirement, but its footer buttons (Confirm and Cancel) can not be removed like the slotted footer in the Dialog.  We need customized buttons, buttons more than confirm and cancel.

### DIFF
--- a/build/bin/new.js
+++ b/build/bin/new.js
@@ -119,7 +119,7 @@ Files.forEach(file => {
 const navConfigFile = require('../../examples/nav.config.json');
 
 Object.keys(navConfigFile).forEach(lang => {
-  let groups = navConfigFile[lang][2].groups;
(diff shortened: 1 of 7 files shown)


Leopoldthecoder:
The commits in this PR have nothing to do with Dialog.

Plus, if you need a service that dynamically creates Dialogs, you can write one in your project. Element itself will provide no such service.

PR #5170, opened 2017-06-02
Locale: Add Latvian translation file.

DavisMiculis:
Added Latvian translation file (lv.js) for locale.


DavisMiculis:
Should the commented `// week` be uncommented too? I just took the default en.js and translated, I left the comments as is.

Leopoldthecoder:
`week` is not used for now, so just leave it. Thanks for contributing.

PR #5461, opened 2017-06-19
Upload: Reset file input value to null after adding files.

elfman:
Fix issue #5455


baiyaaaaa:
input value clear should be do when you click the input again.(like native input file) #5706

PR #5472, opened 2017-06-20
Fix the problem that affects the user experience

leezng:
When the input-number component has the maximum and minimum values, if the value entered is less than the minimum or greater than the maximum, the current input value is adjusted immediately. For example, the minimum value is 100, and I want to enter 300, which is a Legal value, but when I enter the first character "3", it automatically updates the current value of 100, which is very affecting the user experience.


wacky6:
How about adjusting the value on `change` event or `blur` event?

leezng:
Yeah. It will auto adjusting the value on the `blur` event.

PR #5587, opened 2017-06-28
fix carousel bug #5462 && #2189

JoJoChilly:
modified: packages/carousel/src/item.vue
carousel scroll without the end page flash in.

### DIFF
--- a/packages/carousel/src/item.vue
+++ b/packages/carousel/src/item.vue
@@ -12,7 +12,8 @@
     :style="{
       msTransform: `translateX(${ translate }px) scale(${ scale })`,
       webkitTransform: `translateX(${ translate }px) scale(${ scale })`,
-      transform: `translateX(${ translate }px) scale(${ scale })`
+      transform: `translateX(${ translate }px) scale(${ scale })`,
+      zIndex: `${ zIndex }`
     }">
     <div
       v-if="$parent.type === 'card'"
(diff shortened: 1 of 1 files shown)


JoJoChilly:
add: the original has already been fixed in mac, but it still happend on windows browsers like chrome.

Leopoldthecoder:
Unfortunately,  according to my tests, this doesn't work on macOS either. Are you interested in taking a closer look?

JoJoChilly:
Strange because it works in my mac... I will test it on more devices tomorrow.

JoJoChilly:
@Leopoldthecoder Already checked on mac and pro,  works fine. Could you give further information about device type & version and problems?

Leopoldthecoder:
untitled

image

image

As the gif shows, there is a blue flash when the third item is sliding in. Blue is the background color of the last carousel item.

JoJoChilly:
@Leopoldthecoder Got it... I guess it may need a reconfiguration. Will inform you as soon as I finished.

JoJoChilly:
@Leopoldthecoder  Find this better, why not make it inside? https://github.com/ElemeFE/vue-swipe

Leopoldthecoder:
`vue-swipe` is made for mobile devices, so its interactions are different from Carousel.

zbagz:
Same problem here. My content inside the carousel has a background with opacity, so it looks weird when the items are being repositioned.

Maybe we can fix this by adding "transition: 0s" on every element of the carousel and only allow transition time on the active one, his next and his previous item, which are the only ones that should actually be animated and visible on the foreground.

Leopoldthecoder:
Fixed in #6394

PR #5695, opened 2017-07-05
Update index.js

lengjing:


Leopoldthecoder:
Please modify `build/bin/build-entry.js` instead of `src/index.js`, which is automatically generated.

PR #5796, opened 2017-07-11
update element use vue-i18n@6.x and other i18n docs

mario56:

在使用vue-i18n@6.x 时候，pagination total 的i18n使用的是 this.t(key,value)[代码处](https://github.com/ElemeFE/element/blob/dev/packages/pagination/src/pagination.js#L225)

所以如果只用
```
Vue.use(Element, {
  i18n: key => i18n.vm._t(key)
})

```
会丢失pagination total 的值。

### DIFF
--- a/examples/docs/en-US/i18n.md
+++ b/examples/docs/en-US/i18n.md
@@ -102,7 +102,7 @@ const i18n = new VueI18n({
 })
 
 Vue.use(Element, {
-  i18n: key => i18n.t(key)
+  i18n: (key, value) => i18n.t(key. value)
 })
 
 new Vue({ i18n }).$mount('#app')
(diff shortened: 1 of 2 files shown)


johanbove:
Thank you for providing this explanation @mario56 !
I was scratching my head with `total {total}` did not show the actual number...

English translation of the original post by @mario56 for find-ability:

> When using vue-i18n@6.x, the pagination total i18n uses this.t (key, value) code
> So if only used `code block`
> The value of pagination total is lost.

shuidian:
1、here is a problem，when using i18n@6.x，if i write like this：
Vue.use(Element, {
  i18n: (key, value) => **i18n.vm._t(key, value)**
})
the total will always show 'undefined'。
should write like this，
Vue.use(Element, {
  i18n: (key, value) => **i18n.t(key, value)**
})
this will make it right。
2、there is some difference between english i18n.md and chinese i18n.md,
image

PR #5939, opened 2017-07-19
Tag: Reverse closeTransition logic or revise prop name

jameswragg:
`:close-transition="true"` should enable close transition, opposite should also be true.

### DIFF
--- a/packages/tag/src/tag.vue
+++ b/packages/tag/src/tag.vue
@@ -1,5 +1,5 @@
 <template>
-  <transition :name="closeTransition ? '' : 'el-zoom-in-center'">
+  <transition :name="!closeTransition ? '' : 'el-zoom-in-center'">
     <span
       class="el-tag"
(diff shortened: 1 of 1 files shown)


jameswragg:
Alternatively, the prop name could change as currently misleading.
> "close-transition - whether the removal animation is disabled. Default: false"

`:close-transition="false"` leads me to believe there will not be a close animation, where the opposite is true.

Maybe `disable-transitions` would be a better prop name?

Leopoldthecoder:
@jameswragg I think `disable-transitions` is better. Perhaps we can leave `close-transition` for now and change it to `disable-transitions` in the next major version because this is a breaking change.

If you're interested, you can submit the `disable-transitions` pull request to the `carbon` branch. And don't forget the documentations. Thanks.

baiyaaaaa:
That's will be lead to a breaking change, you can provide pr to the carbon branch which is the comming 2.0 version

PR #5989, opened 2017-07-21
optimize code

ProceedCheng:

### DIFF
--- a/packages/tabs/src/tab-nav.vue
+++ b/packages/tabs/src/tab-nav.vue
@@ -113,11 +113,6 @@
         }
       }
     },
-
-    updated() {
-      this.update();
-    },
-
     render(h) {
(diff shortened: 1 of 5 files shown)


packages/tabs/src/tab-nav.vue
```diff
@@ -113,11 +113,6 @@
         }
       }
     },
-
-    updated() {
-      this.update();
```
baiyaaaaa: This will lead to a bug that the nav__active-bar do not scroll to the right place

ProceedCheng: sorry, there is someting wrong with my pr,  what  I'm curious about is that why the update method wrapped in updated function, I mean it could be invoked directly in mounted hooks

PR #6024, opened 2017-07-24
Table: fix hover-row bgcolor of table with fixed column

xtongs:
Fix: The background color of highlighted hover-row or hover-row of table with fixed column do not change when hovering.

### DIFF
--- a/packages/theme-default/src/table.css
+++ b/packages/theme-default/src/table.css
@@ -400,8 +400,14 @@
     }
 
     @e body {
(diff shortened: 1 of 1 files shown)


Leopoldthecoder:
@xtongs Hi, could you write a demo of the bug you're trying to fix?

xtongs:
@Leopoldthecoder Sorry for the absence of demo, but I have recorded a screenshot, check this out.

PR #6101, opened 2017-07-26
RadioGroup:add accessibility

maranran:
add accessibility for RadioGroup:
1、 add some aria role and attrs for screen reader
2、optimize some Keyboard Interaction：
-  The Tab key moves focus between radio button groups and other widgets.
-  Pressing the Shift+Tab keys exits the group and moves focus to the previous form control.
-  When a radio group receives focus:
     If a radio button is checked, focus is set on the checked button.
     If none of the radio buttons are checked, focus is set on the first radio button in the group.

-  Space: checks the focused radio button if it is not already checked.
 
-  Right Arrow and Down Arrow: 
  move focus to the next radio button in the group, uncheck the previously focused button, and check 
  the newly focused button. If focus is on the last button, focus moves to the first button.

-  Left Arrow and Up Arrow: move focus to the previous radio button in the group, uncheck the previously focused button, and check the newly focused button. If focus is on the first button, focus moves to the last button.

### DIFF
--- a/packages/radio/src/radio-button.vue
+++ b/packages/radio/src/radio-button.vue
@@ -6,6 +6,11 @@
       { 'is-active': value === label },
       { 'is-disabled': isDisabled }
     ]"
+    role="radio"
+    :aria-checked="value === label"
+    :aria-disabled="isDisabled"
+    :tabindex="tabIndex"
+    @keydown.space.stop.prevent="value = label"
   >
     <input
       class="el-radio-button__orig-radio"
@@ -13,7 +18,9 @@
       type="radio"
       v-model="value"
       :name="name"
-      :disabled="isDisabled">
+      :disabled="isDisabled"
+      tabindex="-1"
+    >
     <span class="el-radio-button__inner" :style="value === label ? activeStyle : null">
       <slot></slot>
       <template v-if="!$slots.default">{{label}}</template>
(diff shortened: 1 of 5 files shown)


packages/radio/src/radio-button.vue
```diff
@@ -62,6 +69,9 @@
       },
       isDisabled() {
         return this.disabled || this._radioGroup.disabled;
+      },
+      tabIndex() {
+        return !this.isDisabled ? (this._radioGroup ? (this.value === this.label ? 0 : -1) : 0) : -1;
```
baiyaaaaa: 如果是 `_radioGroup` 的话 `this.value` 是一个数组，而 `this.label` 是一个 `string`，这里恒不等于

packages/radio/src/radio.vue
```diff
@@ -79,6 +85,9 @@
         return this.isGroup
           ? this._radioGroup.disabled || this.disabled
           : this.disabled;
+      },
+      tabIndex() {
+        return !this.isDisabled ? (this.isGroup ? (this.model === this.label ? 0 : -1) : 0) : -1;
```
baiyaaaaa: 同上

packages/radio/src/radio-button.vue
maranran: 我看了一下 _radioGroup 的时候 this.value是个值 不是个数组 是我理解的不对嘛 求大佬们给看下

PR #6214, opened 2017-08-02
Picker only emit user change

Dreamacro:
`DatePicker` 系列的 `change` 事件改为仅响应用户交互


wacky6:
I prefer a cleaner approach. Using flags to distinguish user interaction seems unnecessarily complicated.

Dreamacro:
@wacky6 
Emmmm, which approach are you prefer?

wacky6:
I prefer emit change in user-related event handlers, leave out the flags.

wacky6:
@Leopoldthecoder 

I am working on a cleaner picker implementation, with bug fixes and refined `default-value` feature. Should be able to get rid of the `user` flag in this PR.

Leopoldthecoder:
@wack6 Cool

Leopoldthecoder:
@wacky6 In terms of `default-value`, after #7073, range typed DatePicker would break if the initial value is `[]`: https://jsfiddle.net/leopoldthecuber/LkbeLLd9/ . #7233 is a fix. Just a heads-up for your refining it.

wacky6:
Got it.

wacky6:
*-Picker rework mostly finished.

Updating / adding tests. Should be able to PR later this week.

Spoiler: 
* experimental support to bind non-Date value, controlled with `valueFormat`
* fix navigation issue
* refined default-value, add support for array, make it reactive
* refined highlight, can highlight today/current/default separately 
* simplified internal implementation, make room for more features.

wacky6:
May later rework date-table to support Daylight Saving Times.

Currently, it does not highlight properly.

image

Leopoldthecoder:
Wow that's great

PR #6323, opened 2017-08-05
added arabic language locale file

salamaashoush:
Added ar ( Arabic ) language locale file

### DIFF
--- a/src/locale/lang/ar.js
+++ b/src/locale/lang/ar.js
@@ -1,0 +1,104 @@
+export default {
+  el: {
+    colorpicker: {
+      confirm: 'موافق',
+      clear: 'إزالة'
+    },
+    datepicker: {
+      now: 'الآن',
+      today: 'اليوم',
+      cancel: 'إلغاء',
+      clear: 'إزالة',
+      confirm: 'موافق',
+      selectDate: 'أختر التاريخ',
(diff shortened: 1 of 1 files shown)


src/locale/lang/ar.js
```diff
@@ -0,0 +1,104 @@
...
+    tree: {
+      emptyText: 'لايوجد بيانات'
+    },
+    transfer: {
+      noMatch: 'لايوجد بيانات مطابقة',
+      noData: 'لايوجد بيانات',
+      titles: ['List 1', 'List 2'], // to be translated
+      filterPlaceholder: 'Enter keyword', // to be translated
+      noCheckedFormat: '{total} items', // to be translated
+      hasCheckedFormat: '{checked}/{total} checked' // to be translated
```
Leopoldthecoder: These four lines also need to be translated.

Leopoldthecoder:
Thanks for contributing.

Off the topic, what is the difference between Arabic and [Farsi](https://github.com/ElemeFE/element/blob/dev/src/locale/lang/fa.js)? Forgive me if I'm asking a stupid question.

salamaashoush:
Arabic and Farsi(Persian) , They use the same Arabic alphabet ,but sound completely different

Issue #6440, opened 2017-08-12
[Bug Report] el-col: <el-col :span="0"> cant't get the 'el-col-0' class name


laomao800:
### Element UI version
1.4.2

### OS/Browsers version
mac chrome

### Vue version
2.4.2

### Reproduction Link
[https://jsfiddle.net/tkz7rkry/](https://jsfiddle.net/tkz7rkry/)

### Steps to reproduce
```
<el-row>
  <el-col :span="0"></el-col>
</el-row>
```

### What is Expected?
get `<div class="el-col el-col-0">`

### What is actually happening?
only get `<div class="el-col">`

Leopoldthecoder:
Fixed in 2.0.

PR #6441, opened 2017-08-12
Col: fix <el-col :span="0"> cant't get the 'el-col-0' class name

laomao800:
Fixed https://github.com/ElemeFE/element/issues/6440

`<el-col :span="0"></el-col>` expect to get `<div class="el-col el-col-0">` but only get the `<div class="el-col">`


Leopoldthecoder:
Fixed in 2.0.

PR #6446, opened 2017-08-13
Table: fix  sort table bug #6444

Weiting-Zhang:
fix sort table bug #6444


Weiting-Zhang:
Technically it's not a bug, since the website says 

>  sorting method works when sortable is true. Should return a boolean.

But I think maybe it's more convenient and reasonable to accept both boolean return  value and number return value to sort the table ?

furybean:
This PR has conflicts currently.
In my test this bug has already fixed in version 2.0.11: https://jsfiddle.net/ngdwetb0/2/

PR #6569, opened 2017-08-18
Carousel: fix z-index bug

thunderrun:
Inactive carousel items' `z-index` are all `0`
this result some items may incorrectly covering the previous active item during the transition.

In the example below the third item `e.g. translateX(255)` sometimes covering the first item `e.g. translateX(-255)` during the transition when the second item is active.
https://jsfiddle.net/po0utg8n/1/


Leopoldthecoder:
Thanks for contributing. This is actually fixed in #6394

PR #6768, opened 2017-08-30
DatePicker:fix month status check Bug

qingdengyue:
DatePicker:fix month status check Bug
There are basic rules:
1.For **every day of every month** ,if the 'disabledDate' function's **result is false**.Means that the Month can't be disabled.
2.For **any day of every month** ,if the 'disabledDate' function's **result is false**.Means that the Month can't be disabled.
   Examples:
   Like '2017-08-30'.The Month is 08
   For '2017-08-29' , the result of 'disabledDate' function is true
   For '2017-08-28' , the result of 'disabledDate' function is false
   At last,the Month 08 should be not disabled.
   
3.For **every day of every month** ,if the 'disabledDate' function's **result is true**.Means that the Month must be disabled.

#6764

### DIFF
--- a/packages/date-picker/src/basic/month-table.vue
+++ b/packages/date-picker/src/basic/month-table.vue
@@ -78,17 +78,11 @@
           while (date < nextMonth) {
             if (this.disabledDate(date)) {
               date = new Date(date.getTime() + 8.64e7);
+              flag = true;
             } else {
+              flag = false;
               break;
             }
-          }
-          // There is a bug of Chrome.
-          // For example:
-          // var date = new Date('1988-04-01 00:00:00') Fri Apr 01 1988 00:00:00 GMT+0800 (CST)
-          // date.setMonth(4) Sun May 01 1988 00:00:00 GMT+0900 (CDT)
-          // Sometimes the time zone will change.
-          if (date - nextMonth < 8.64e7) {
-            flag = true;
           }
         }
 


lhajh:
@Leopoldthecoder 没看懂，意思是你们早就修复好了吗，那为什么现在你们官网上面还是有这个问题

Leopoldthecoder:
@lhajh 昨天刚合并，还没发布。

lhajh:
@Leopoldthecoder 好的，请尽快发布吧，这个问题有点大啊。还有，非常感谢你们

georgehpj:
year status has the same bug, will it be fixed as well?

wjf444128852:
昨天也遇到了，之前都是好好的，現在需要添加個別新的功能，於是把開發包挪到新的文件夾重新install的結果出現了這個 BUG。

remzmike:
There's still a bug here in 1.4.4 as far as I can tell and I think this is part of the reason:

```javascript
d = new Date(0);
d.setFullYear(2017);
d.setMonth(8);
// > Sun Oct 01 2017 19:00:00 GMT-0400 (Eastern Daylight Time)
d.setMonth(8);
// > Fri Sep 01 2017 19:00:00 GMT-0400 (Eastern Daylight Time)
```

Fix options:
1) Fix it using jank javascript dates and lots of testing
2) Use moment.js
3) Expose this functionality differently (eg. expose getCellStyle instead)

This happens in Chrome and Firefox.

remzmike:
Probably because of this: https://stackoverflow.com/questions/35822352/date-class-setmonth-sets-weird-values

remzmike:
This might happen because `new Date(0)` has a date component of 31, which not all months have.

`new Date(year, month, 1, 0)` doesn't show the same problem.

However, the `nextMonth` calculation still depends on the `setMonth` and will probably still be buggy because of it.

remzmike:
nextMonthOrdinal = (month + 1) % 12

remzmike:
Here's a similar fix for the year table.

```javascript
var date = new Date(year, 0, 1, 0);
var nextYear = new Date(year + 1, 0, 1, 0);
```

https://gist.github.com/remzmike/a06aaa7a8460b3c58f199b4bc68e9ef7

PR #6837, opened 2017-09-04
Select: watch old value on select change

junjunconan:
In some cases I need to get old value on select change. For example, If I have a group of select components which share the same option list (let's say ['a', 'b', 'c']). When I select 'a' in the first select component, I want to filter out 'a' in the option list and make the option list only ['b', 'c']. After that when I change the selection from 'a' to 'b', I want to filter out 'b' and put 'a' back to the list and the list will then be ['a', 'c'].

In this case, I need to know the old value I previously selected and put it back to the option list.

### DIFF
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -230,7 +230,7 @@
         this.cachedPlaceHolder = this.currentPlaceholder = val;
       },
 
-      value(val) {
+      value(val, oldVal) {
         if (this.multiple) {
           this.resetInputHeight();
           if (val.length > 0 || (this.$refs.input && this.query !== '')) {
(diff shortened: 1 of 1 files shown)


wacky6:
You don't have to modify change event for this. https://jsfiddle.net/e408afn9/

This PR will make select's change inconsistent with others.

junjunconan:
What if I have multiple select fields which share the same option list, or even I want to dynamically add or remove fields? For example I have 3 fields, I select 'a' in the first field and select 'b' in the second, but I don't want to have option 'a' and 'b' appeared in the third field.

wacky6:
Have a look at computed property. https://vuejs.org/v2/guide/computed.html

junjunconan:
OK, got it. Thanks.

PR #6871, opened 2017-09-05
fix code readability and performance optimization for year and month

huangdanfeng:
By comparing the last day of a year or a month, you can verify that the time button is disabled and no excessive loops are needed, so that performance and code readability can be improved.


Leopoldthecoder:
What if only part of a month is disabled? For example, if we only disable 2017/09/25 - 2017/09/30, comparing the last day (2017/09/30) will make the whole month disabled. So we have to compare each day of a month.

Issue #6905, opened 2017-09-07
[Bug Report] Component `select` won't auto collapse when click outside of this component


reclay:
### Element UI version
1.4.4

### OS/Browsers version
chrome 60

### Vue version
2.4.0

### Reproduction Link
[https://jsfiddle.net/793n0uLe/1/](https://jsfiddle.net/793n0uLe/1/)

### Steps to reproduce
Steps to reproduce
1. Click the third `select` component: this component is expanded
2. Click outside of this component:  this component doesn't collapse.

This issue is caused by `nodeList`(in `clickoutside.js`) deleting incorrect item, item which to be deleted shouldn't determined by `id`.
  
![image](https://user-images.githubusercontent.com/14367118/30151316-892d7476-93e1-11e7-9b51-14ce28eaf965.png)

### What is Expected?
This `select` component will auto collapse.

### What is actually happening?
This `select` component doesn't collapse.

Leopoldthecoder:
Fixed in #6973

PR #6907, opened 2017-09-07
select: fix collapse bug

reclay:
issue: [https://github.com/ElemeFE/element/issues/6905](https://github.com/ElemeFE/element/issues/6905)


Leopoldthecoder:
#6973

Issue #6924, opened 2017-09-08
[Bug Report] Table column width is not computed correctly when `scrollY` is false


reclay:
### Element UI version
1.4.4

### OS/Browsers version
chrome 61

### Vue version
2.4.0

### Reproduction Link
[https://jsfiddle.net/m8ch4vhj/](https://jsfiddle.net/m8ch4vhj/)

### Steps to reproduce
This issue is caused by `table-layout.js` which did not take into account the situation of `scrollY`
![image](https://user-images.githubusercontent.com/14367118/30194836-2681ffe8-9488-11e7-804d-aaa14526f4d8.png)

### What is Expected?
    let totalFlexWidth = bodyWidth - bodyMinWidth;
    totalFlexWidth -= this.scrollY ? this.gutterWidth : 0;

### What is actually happening?
    const totalFlexWidth = bodyWidth - this.gutterWidth - bodyMinWidth;

liril-net:
Fixed in [Carbon](https://github.com/ElemeFE/element/releases/tag/v2.0.0)

PR #6926, opened 2017-09-08
table: fix table width bug

reclay:
issue: [https://github.com/ElemeFE/element/issues/6924](https://github.com/ElemeFE/element/issues/6924)
issue: #3708


Leopoldthecoder:
Fixed in 2.0

PR #6989, opened 2017-09-13
doc: Add scheme part of URL in import via CDN examples

yoichi:
When I tried locale setting with element-ui imported via CDN, sample code didn't work since scheme part of URL is missing.

Thanks,


Leopoldthecoder:
It'll work once you deploy it on a server and visit it from a browser.

yoichi:
Does i18n.md have prerequisites other than installation.md? In the sample code of installation.md, "https:" is included in the URL.

Leopoldthecoder:
`https:` is not necessary.

PR #7056, opened 2017-09-15
Add enter event for "input" component

noomerzx:
I've alway use input component for any search bar and it's not support the "Enter" event and a lot of user use that event to submit search, So i've added the support for enter event to "Input" component.


Leopoldthecoder:
https://jsfiddle.net/wk9c4fhm/

PR #7196, opened 2017-09-25
Accessibility for Input & autocomplete & input-number & progress & Rate & Upload & Switch & collapse

maranran:
add Accessibility for Input & autocomplete & input-number & progress & Rate & Upload & Switch & collapse including aria roles & props &states and keyboard access
Detail:
Input-number:
1、add role & state & prop 
2、can be tab to  and  use up/down/left/right to change value 
Autocomplete:
1、suggestion list will expend when input focus and then you can use up/down cross suggestion list ,use enter key to select one 
2、add role & state & prop 
Progress: 
add role & state & prop
Rate:
1、 can be tab to  and use  up/down/left/right to change value 
2、add role & state & prop
Upload：
1、 can be tab to  and use  delete  file list
2、operation tips will show only when focus
Switch:
1、add role & state & prop
2、 can be tab to  and use enter change value
Collapse：
1、add role & state & prop
2、can be tab to  and use enter expend the tab panel


abhipanda:
Thanks for doing this - the only thing stopping us from using element library was accessibility and if this remains our focus then kudos !

maranran:
Thanks for your attention, accessibility for other components are in progress， I think we can do it together  @abhipanda @Leopoldthecoder

PR #7275, opened 2017-09-28
Form-item: fix "required" attribute is not effective for validation

HoxtonL:

### DIFF
--- a/packages/form/src/form-item.vue
+++ b/packages/form/src/form-item.vue
@@ -148,7 +148,7 @@
     methods: {
       validate(trigger, callback = noop) {
         var rules = this.getFilteredRule(trigger);
-        if (!rules || rules.length === 0) {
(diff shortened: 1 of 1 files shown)


Leopoldthecoder:
Could you write a demo to reproduce this bug?

HoxtonL:
Here is the link: https://codepen.io/anon/pen/yzXLRE?editors=1010
When 'isRequired' is changed, it can only change class of the 'form-item' component, but not its validate rules.
That is because in the 'mounted' hook, it can't not create a listen for event 'el.form.blur' or 'el.form.change' when 'getRules' returns an empty array no matter it has a 'required' attribute or not.
But even if you add 'rules' in the form-item component, the 'required' attribute still can not effect the validate rules. Because it doesn't look for the 'required' attribute while building rules for validation.

Leopoldthecoder:
Thanks.

PR #7387, opened 2017-10-09
Table: add server side sort feature

yugasun:
Add server side sort parameter `ajax-sort` for component `Table`。

### DIFF
--- a/examples/docs/en-US/table.md
+++ b/examples/docs/en-US/table.md
@@ -1,7 +1,10 @@
 <script>
+  import { orderBy } from '../../../packages/table/src/util';
+
   export default {
     data() {
       return {
+        loading: false,
         tableData: [{
           date: '2016-05-03',
           name: 'Tom',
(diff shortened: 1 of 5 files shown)


marcelogarbin:
@yugasun Hello, would you have any tutorial on how to implement server side in the table?

yugasun:
@marcelogarbin Here is a simple demo for you: https://jsfiddle.net/yugasun/7kh2nrj5/, wish can help you.

marcelogarbin:
@yugasun Thanks, but I can do this already. What I can not do is to add action buttons on the table, so I click to go to the edit page of the registry. As in the image below (it's [Element](http://element.eleme.io/#/en-US/component/table) table)::

image


I tried to make, but without success, the error or it does not work, I could not find any examples that I could follow and work.

I read something about `cellRenderer`, but I have no way of doing it =/

Would you have something about it? add buttons with click function...

PR #7480, opened 2017-10-13
Carbon: date-picker's two panels could pick date alone

waynecz:


wacky6:
I suggest write some tests, otherwise it might get broken by someone else after some time.

I prefer to keep existing date-picker appearance, and only show left/right arrow if `unlock-panel-link` is true.

May be a better prop name? `unlink-panels` ???

wacky6:
See also https://github.com/ElemeFE/element/issues/6935

@renxia Any suggestions / opinions?

PR #7590, opened 2017-10-19
Carbon: Table add features and fix bugs

liril-net:

### DIFF
--- a/packages/table/src/table-body.js
+++ b/packages/table/src/table-body.js
@@ -55,16 +55,74 @@ export default {
                 on-mouseleave={ _ => this.handleMouseLeave() }
                 class={ [this.getRowClass(row, $index)] }>
                 {
-                  this._l(this.columns, (column, cellIndex) =>
-                    <td
-                      class={ [column.id, column.align, column.className || '', columnsHidden[cellIndex] ? 'is-hidden' : '' ] }
-                      on-mouseenter={ ($event) => this.handleCellMouseEnter($event, row) }
-                      on-mouseleave={ this.handleCellMouseLeave }>
-                      {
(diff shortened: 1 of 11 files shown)


packages/table/src/table-body.js
```diff
@@ -79,8 +137,6 @@ export default {
                 : ''
               ]
             ).concat(
-              this._self.$parent.$slots.append
```
syn-zeta: @liril-net @Leopoldthecoder Any particular reason this slot was moved? It breaks the infinite loading plugins for the table: https://jsfiddle.net/4ez4j2j4/ (try removing the `@next` in HTML section line 2 to see it working).

packages/table/src/table.vue
```diff
@@ -39,6 +40,9 @@
       <div :style="{ width: bodyWidth }" class="el-table__empty-block" v-if="!data || data.length === 0">
         <span class="el-table__empty-text"><slot name="empty">{{ emptyText || t('el.table.emptyText') }}</slot></span>
       </div>
+      <div class="el-table__append-wrapper" ref="appendWrapper" v-if="$slots.append">
+        <slot name="append"></slot>
+      </div>
```
Leopoldthecoder: @syn-zeta This slot is moved outside the table body, so the content shouldn't be nested in `<tr>` and `<td>`:https://jsfiddle.net/4ez4j2j4/1/ . Looks we should address this in the breaking change @liril-net

packages/table/src/table-body.js
liril-net: `append` slot should be unique in one `el-table` to not affect `render` in `el-tbody`, so i move it out of `el-tbody`.

PR #7765, opened 2017-10-26
Remove implicit any in type definition

Raganof:
Make the definitions compliant with the option `--noImplicitAny` of [typescript's compiler](https://www.typescriptlang.org/docs/handbook/compiler-options.html).

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -77,7 +77,7 @@ declare namespace ElementUI {
    * Please do not invoke this method directly.
    * Call `Vue.use(ElementUI)` to install.
    */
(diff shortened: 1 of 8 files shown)


types/index.d.ts
```diff
@@ -283,4 +283,4 @@ declare namespace ElementUI {
   export class Upload extends ElUpload {}
 }
 
-export default ElementUI
+export = ElementUI
```
Leopoldthecoder: Hi, could you please tell me why this line is changed?

Raganof: Hi, this was a mistake, I'm removing this change.

PR #7798, opened 2017-10-27
Eslint: fix eslint config for deploy build

yugasun:
- [x] Update eslint config for deploy build scripts
- [x] Fix deploy build eslint errors


Leopoldthecoder:
It's not necessary to list these packages because they are dependencies of `cooking-lint`. When you run `yarn`, they'll be installed as well.

PR #7840, opened 2017-10-30
input-number allow input '.' '-'

linzer:
input-number allow input '.' '-'


Leopoldthecoder:
Any reason we should allow `-`?

linzer:
can't input negative numbers now

PR #7865, opened 2017-10-30
Select: Fix size el-tag in multiple select

alexdee2007:
Size el-tag in multiple select is always small.
This change fix size el-tags as selectSize.


Leopoldthecoder:
#7872

PR #3910, opened 2017-03-30
Add TypeScript definitions

hjkcai:
This is a WIP PR but I want to post it before it completes because **I am not sure my definitions 100% match the API**. 

Every definition comes from the English version in the official documentation [http://element.eleme.io/](http://element.eleme.io/). But the documentation is too ambiguous to get a detailed definition. So I think more discussions are needed!

Please also help me to solve the 'todos' (lines start with `// TODO`). Those todos are the problems I have encountered so far.

TODOs left:

- [x] `date-picker.d.ts` Description of `DateRange`
- [x] `select.d.ts` Use a better interface name of `AlterItemsEventHandler`
- [ ] `table-column.d.ts` Complete type `TableColumn` and `ElTableColumn.filteredValue`
- [ ] `tree.d.ts` Complete type and description of `FilterNodeMethod`
- [ ] `upload.d.ts` Full review needed

PR #7939, opened 2017-11-01
Fix inconsistency of the TypeScript definitions - Solution 1

hjkcai:
@vtroop asked a question about importing element-ui in TypeScript in #3910 today. And I notice that there is a critical inconsistency in index.d.ts:

```typescript
declare namespace ElementUI { ... }
export default ElementUI
```

When importing element-ui, TypeScript will do conversions like this:

```typescript
import ElementUI from 'element-ui'
import * as ElementUI from 'element-ui'
```

```javascript
var ElementUI = require('element-ui').default
var ElementUI = require('element-ui')
```

But Babel and webpack will try to find default export for compatibility like this:

```javascript
import ElementUI from 'element-ui'
// -->
var ElementUI = require('element-ui').default || require('element-ui')
```

In the source file, components and functions are exported via `module.exports`. There is no `default` entry in the exports.

The first solution is the same as this PR does: to change the definition file and tell users to import element-ui like this:

```typescript
import * as ElementUI from 'element-ui'
```

See #7940 for solution 2.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -283,4 +283,4 @@ declare namespace ElementUI {
   export class Upload extends ElUpload {}
 }
 
-export default ElementUI
+export = ElementUI


hjkcai:
@Leopoldthecoder

PR #7940, opened 2017-11-01
Fix inconsistency of the TypeScript definitions - Solution 2

hjkcai:
See descriptions in #7939

The second Solution is to add a `default` entry in the exports. So you can keep the original typings and users can import element-ui like this:

```typescript
import ElementUI from 'element-ui'
```

Then TypeScript will resolve the module correctly.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -248,3 +248,5 @@ module.exports = {
   Main,
   Footer
 };
+
+module.exports.default = module.exports;


Leopoldthecoder:
Thank you so much for the clarification. I'll go for this solution.

But `src/index.js` is automatically generated by `build/bin/build-entry.js` (see the first line of src/index.js), so we shouldn't edit it directly. Could you edit `build/bin/build-entry.js` instead?

vtroop:
Thanks for your work again~
So, the problem is due to the namespace of typescript, is it?

hjkcai:
Sure. I will open another PR.

hjkcai:
@vtroop 这个锅ts不背😂  主要是webpack、babel和ts在模块解析上有不同的地方。webpack和babel有照顾node模块的写法（看 #7939 ），但ts没照顾，用es6 import就全要用es6 import

懒得写英文了233333 （逃

vtroop:
@hjkcai I see, the experience of learing `import` in javascript makes me confused. The better way is to use es6 import. I'm a new learner in ts, 2333

hjkcai:
@vtroop 😂 加油加油…不过如果你没说的话还发现不了这个问题呢，这个问题的解决还得归功于你哈

vtroop:
Hi @hjkcai @Leopoldthecoder 
I'm sorry to bother you again.
This solution can't solve this problem such as 
```javascript
import {Message} from 'element-ui'
```
As far as i'm concerned, we need to edit like 
```javascript
//index.d.ts
export = ElementUI
```
And in this way, there will be no problem.

hjkcai:
@vtroop See #8223

PR #7973, opened 2017-11-02
【Accessibility】 Cascader &  Dropdown

maranran:
dropdown
1、you can use tab to focus trigger element; use enter key open the  dropmenu and esc close it ; use up/down focus through dropmenu and enter key select menuitem and close it; and you can focus the trigger element next time you refocus this component use tab;

2、add aria role and attr for trigger element and dropmenu
cascader
1、you can use tab to focus input element; use enter key open the  c and esc close it ; use left/right/up/down focus through menu and enter key select menuitem and close it;
2、add aria role and attr for cascader


maranran:
这个单元测试是不是 setTimeout后遗症 时不时的就挂了？？

PR #8009, opened 2017-11-03
【accessibility】popover & tooltip & message & Notification

maranran:
popover & tooltip
1、when you  focus the trigger element , the tooltip will show and  notice the AT
 message & Notification
1、when you  focus the trigger element , press enter key the message will show and notice the AT


maranran:
nodeType  bug已修复

PR #8010, opened 2017-11-03
Cascader: optimize hover to expand experience

Dzinlife:
group

When cursor moves between menus, the submenus switch immediately, it is hard to select the desired item.

A better way is to include a bit of a delay when activating submenus, but it will make the menu switch slowly.

The perfect UX is to keep menus switch fast, only hold it when we want to select the submenus. 
Here is a solution:

group

Idea from OSX. The red area prevent menus switch for a short time.

### DIFF
--- a/packages/cascader/src/menu.vue
+++ b/packages/cascader/src/menu.vue
@@ -39,7 +39,8 @@
         value: [],
         expandTrigger: 'click',
         changeOnSelect: false,
-        popperClass: ''
+        popperClass: '',
+        hoverTimer: 0
       };
     },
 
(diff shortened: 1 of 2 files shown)


packages/cascader/src/menu.vue
```diff
@@ -138,6 +138,33 @@
         popperClass
       } = this;
 
+      let hoverMenuRefs = {};
+      this.hoverTimer = null;
```
Leopoldthecoder: `this.hoverTimer` 最好在 data 里注册一下

packages/cascader/src/menu.vue
```diff
@@ -138,6 +138,33 @@
...
+      const hoverMenuHandler = e => {
+        const hoverHreshold = 500;
```
Leopoldthecoder: `hoverThreshold`

Leopoldthecoder:
最好自己 fork 一下项目，从自己项目的分支提 PR，这样能避免在主仓库直接创建分支。

Dzinlife:
嗯嗯好的

PR #8141, opened 2017-11-09
[Form]: added validateOnRuleChange

badpunman:
In my use case, the rules provided to form change dynamically. The form validating itself every time is not welcome.

I would add this validateOnRuleChange boolean value, that can be changed to false to prevent this behavior.

Also the watcher didn't provide a callback to the validate() so the rejected promise escaped to console and was shown as error.

### DIFF
--- a/packages/form/src/form.vue
+++ b/packages/form/src/form.vue
@@ -34,11 +34,17 @@
         type: Boolean,
         default: true
       },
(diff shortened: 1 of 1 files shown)


badpunman:
I suppose I'll have to use a personal branch of the library if this won't get implemented.

Issue #7465, opened 2017-10-13
[Bug Report] Memory leak caused by el-select


STLighter:
### Element UI version
1.4.6

### OS/Browsers version
Windows 7 x64/ Chorme v61.0.3163.100 x64

### Vue version
2.5.0 (not only)

### Reproduction Link
Momery leak when using el-select: [https://jsfiddle.net/ST_Lighter/tev5L23h/](https://jsfiddle.net/ST_Lighter/tev5L23h/)
No leak when using el-input: [https://jsfiddle.net/ST_Lighter/19uthaf9/](https://jsfiddle.net/ST_Lighter/19uthaf9/)

### Steps to reproduce
Click run button will start a interval. It will update a counter.
Then check the heap snapshot in develop tools ( click gc button before taking snapshot).
( In the 1st link I use el-select and in the 2nd I use el-input, you will find different result)

### What is Expected?
No memory leak in both link

### What is actually happening?
When the code using el-select ( the 1st link) running, the memory using will increase continuing
![using el-select](https://user-images.githubusercontent.com/10627202/31529541-7a1db670-b00c-11e7-89fe-bfbb4a7bcdaa.png)
It not happend when using el-input.
![using el-input](https://user-images.githubusercontent.com/10627202/31529570-97dde536-b00c-11e7-8321-0c902e97fd53.png)

badpunman:
These mostly seem to be cleared up when the component is destroyed. I'm however concerned over what is left after the component is destroyed. There seems to be something related to el-scrollbar element that is leaving vnodes allocated.
![image](https://user-images.githubusercontent.com/937289/32666707-01b1bf42-c641-11e7-92e2-3ef6da753dc1.png)

There is something suspect with the way mouseUpDocumentHandler has $parent scrollbar/main.vue in its context and all. The way this renders and then shares reference to wrap via computed properties is a pattern i'm not familiar with.

badpunman:
Oops, upon further inspection:

These nodes were only kept alive because scrollbar's mouseUpDocumentHandler had itself as context indeed, but since this is a cyclic reference only, browser was able to release them after instructing it to manually do a garbage collection. It's still a bit bad to have such a reference going on though. Not sure how to fix it however.

badpunman:
However, I think I've identified what does cause some leakage here in your scenario, and I had it fixed in my own code while looking at other possible leaks.
[...]

PR #8168, opened 2017-11-10
clickoutside: fix memory leak

badpunman:
found while investigating https://github.com/ElemeFE/element/issues/7465

clickoutside directive didn't clean up properties it added to elements when they were bound. 

this added unnecessary function references to element's own methods that prevented garbage collection.


Leopoldthecoder:
@badpunman Thanks

PR #8225, opened 2017-11-13
clickoutside: fix memory leak

STLighter:
Relative issue: [#7465](https://github.com/ElemeFE/element/issues/7465)

`documentHandler` use an old vnode in closure, which caused unexpected memory leak.


badpunman:
Indeed. Makes some sense now.

PR #8260, opened 2017-11-14
Dialog: Add transition name as a prop (with default value for reverse compatibility)

SabatinoMasala:
I added the transition name as a prop on the dialog component, with a default value of 'dialog-fade'. This way we can fully control the transition of the dialog.

### DIFF
--- a/packages/dialog/src/component.vue
+++ b/packages/dialog/src/component.vue
@@ -2,20 +2,20 @@
   <transition :name="transition">
     <div class="el-dialog__wrapper" v-show="visible" @click.self="handleWrapperClick">
       <div
-              class="el-dialog"
-              :class="[{ 'is-fullscreen': fullscreen, 'el-dialog--center': center }, customClass]"
-              ref="dialog"
-              :style="style">
+        class="el-dialog"
+        :class="[{ 'is-fullscreen': fullscreen, 'el-dialog--center': center }, customClass]"
+        ref="dialog"
+        :style="style">
         <div class="el-dialog__header">
           <slot name="title">
             <span class="el-dialog__title">{{ title }}</span>
           </slot>
           <button
-                  type="button"
-                  class="el-dialog__headerbtn"
-                  aria-label="Close"
-                  v-if="showClose"
-                  @click="handleClose">
+            type="button"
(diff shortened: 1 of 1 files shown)


Leopoldthecoder:
Hi, thanks for contributing.

Customizing Dialog animation does not seem to be a common request, so we may not add a prop to do that. Actually you can do so by overriding its CSS: https://jsfiddle.net/qs0cod48/

SabatinoMasala:
@Leopoldthecoder, agreed, but my main concern was about being able to disable the transition.

[Other components](https://github.com/ElemeFE/element/blob/dev/packages/tag/src/tag.vue#L2) offer a 'disableTransitions' prop, I could send a PR with an implementation in the dialog, if that fits better? I agree that this would be more consistent than adding the transition name as a prop.

Leopoldthecoder:
Disabling transitions is actually something we can work on. Maybe we'll try to implement it so that
```JS
Vue.use(Element, { transitions: false })
```
can turn off all popup transitions (Dialog, MessageBox, Select, Dropdown, etc.).

For now you can disable Dialog's transition with CSS: https://jsfiddle.net/qs0cod48/1/ . Note that I also disabled backdrop's transition as well in the demo, otherwise it would look kinda weird.

SabatinoMasala:
Sounds very interesting, so we'd store the `transitions` bool on the ELEMENT config variable...
```js
...

ELEMENT.transitions = opts.transitions || true;

Vue.prototype.$ELEMENT = ELEMENT;
```

...and then in components we'd check whether or not to add a transition name:

```vue
<transition :name="$ELEMENT.transitions ? 'el-zoom-in-center' : ''">
  ...
</transition>
```

Seems a bit cleaner than overriding the CSS imo, but happy to hear your thoughts!

A note about reverse compatibility: Would you keep the `disableTransitions` prop on the Tag for example, or strip it out entirely?

Leopoldthecoder:
Yes, we'll keep Tag's `disableTransitions`. And it should have a higher priority than `$ELEMENT.transitions`.

SabatinoMasala:
Great, I will send a PR!

PR #8353, opened 2017-11-17
make pagination responsible to the URL param

zzzgit:
1. if the current page of the pagination changes, the corresponding param will change automatically
2. when mounted, if there is a param in URL, it will be the default value of current page
3. all the above functionality only are available when `token` is set in the `el-pagination` tag

### DIFF
--- a/packages/pagination/src/pagination.js
+++ b/packages/pagination/src/pagination.js
@@ -40,7 +40,9 @@ export default {
 
     prevText: String,
 
-    nextText: String
+    nextText: String,
+
+    token: String
   },
 
   data() {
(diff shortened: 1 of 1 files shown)


wacky6:
I think this feature should be done outside of the component. Pagination is not meant to change URL.

Leopoldthecoder:
I agree with @wacky6. 

This feature is more suitable to be implemented in a wrapper component, which reflects individual project's situation.

zzzgit:
okay i agree with basically
but i don't think it's easy to implement a wrapper which extend el-pagination and has the functionality i mentioned above

wacky6:
Wrapper should not be that difficult, you can watch related events then change urls.

Or, if you want to get it done quickly (and dirty probably?), use vue's $watch API to monitor pagination's internal states.

PR #8428, opened 2017-11-22
[Loading]: .close() can be called earlier than visible=true

koresar:
TL;DR: Sometimes the async sequence of operations can be: `this.visible = false; this.visible = true;` leaving the "Loading" opened forever.

In some scenarios the `.close()` can be called before
```
  Vue.nextTick(() => {
    instance.visible = true;
  });
```
Scenarios include some in-memory caching (Relay/Apollo GraphQL for example).

First time the "Loading" div shows fine. The second time (when a value is cached in memory) it shows and stays forever because `.close()` was called right after the `Loading` constructor call.

The proposed fix solves the bug. Not sure if that's the right way, or if I broke anything else. Probably need testing from core maintainers.

Happy to rework if needed.

### DIFF
--- a/packages/loading/src/index.js
+++ b/packages/loading/src/index.js
@@ -19,21 +19,23 @@ LoadingConstructor.prototype.originalPosition = '';
 LoadingConstructor.prototype.originalOverflow = '';
 
 LoadingConstructor.prototype.close = function() {
-  if (this.fullscreen) {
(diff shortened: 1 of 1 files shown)


wacky6:
Recommend add tests for your use case.
Also, have a look at the failed test.

PR #8440, opened 2017-11-22
remove unnecessary event

njleonzhang:
The event emit is not necessary according to my understanding.


njleonzhang:
@Leopoldthecoder 老哥还是胆子大。

这个PR, 有没有说法啊 https://github.com/ElemeFE/element/pull/6740

PR #8506, opened 2017-11-26
Allow to create new items using keyboard only

lukaszb:
IMHO components like that should work without need to use mouse or having to use arrow keys to hover before typing enter. With this change user would be able to simply type a value in and hit enter to add it. It's way more convenient in my view.

### DIFF
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -615,8 +615,17 @@
       },
 
       selectOption() {
+        const visibleOptions = this.options.filter(option => option.visible);
         if (this.options[this.hoverIndex]) {
           this.handleOptionSelect(this.options[this.hoverIndex]);
+        } else if (visibleOptions.length) {
+          const newOption = this.getNewOption();
+          if (newOption && newOption.value.length > 0) {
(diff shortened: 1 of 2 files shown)


xeoshow:
Exactly the feature what I need, thanks!

marceloavf:
Great feature!

lukaszb:
@Leopoldthecoder any chances you can review this and tell if it's acceptable?

lukaszb:
@Leopoldthecoder any chances you can review this and tell if it's acceptable?

Leopoldthecoder:
For this purpose, you can use `default-first-option`: https://jsfiddle.net/ma106uxt/ .

There is a minor unexpected behavior though: https://jsfiddle.net/ma106uxt/1/ .When you type `'h'`, Select highlights `HTML` instead of the newly created `h`. I have fixed this in #8790 to ensure the created option always gets highlighted before other options.

lukaszb:
@Leopoldthecoder cool, thanks! Totally missed that option. Do you consider adding it to the examples?

Leopoldthecoder:
@lukaszb Sure.

PR #8703, opened 2017-12-06
[DateTimePicker] Adds start, end and step time select attributes

pushchris:
Currently a TimePicker is able to have the start time, end time and step set through the ```picker-options``` but this does not carry through to DateTimePicker's. This pull request adds these attributes in.


wacky6:
It does not work this way.
These options are for time-select only. They do not work with time-picker which is used in date-time-picker.

PR #8940, opened 2017-12-21
[table] Wrong with immutable when use fixed columns(fix immutable updateColumns)

kolesoffac:
### Element UI version
2.0.8

### OS/Browsers version
macos / chrome 63.0.3239.84

### Vue version
2.5.13

### Reproduction Link
[https://jsfiddle.net/8yLhv5cq/](https://jsfiddle.net/8yLhv5cq/)

### Steps to reproduce
- click button
- result: hide content column

### What is Expected?
Not hide content fixed column

Hide content fixed column

https://github.com/ElemeFE/element/blob/0d952b79174520957e86dc5f4bf6cccd96a58ca0/packages/table/src/table-store.js#L332
When change all item in object then computed prop _leftFixedLeafCount_
https://github.com/ElemeFE/element/blob/0d952b79174520957e86dc5f4bf6cccd96a58ca0/packages/table/src/table-body.js#L186
don't recalculate because object _state_ not [change](https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats).

So need return new object. Or use get methods instead computed prop.

Thanks

### DIFF
--- a/packages/table/src/table-store.js
+++ b/packages/table/src/table-store.js
@@ -352,6 +352,8 @@ TableStore.prototype.updateColumns = function() {
 
   states.columns = [].concat(fixedLeafColumns).concat(leafColumns).concat(rightFixedLeafColumns);
   states.isComplex = states.fixedColumns.length > 0 || states.rightFixedColumns.length > 0;
+
+  this.states = {...states};
 };
 
 TableStore.prototype.isSelected = function(row) {
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": [["es2015", { "loose": true }]],
-  "plugins": ["transform-vue-jsx"],
+  "plugins": ["transform-vue-jsx", "transform-object-rest-spread"],
   "env": {
     "utils": {
       "plugins": [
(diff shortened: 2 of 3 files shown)


packages/table/src/table-store.js
```diff
@@ -352,6 +352,8 @@ TableStore.prototype.updateColumns = function() {
 
   states.columns = [].concat(fixedLeafColumns).concat(leafColumns).concat(rightFixedLeafColumns);
   states.isComplex = states.fixedColumns.length > 0 || states.rightFixedColumns.length > 0;
+
+  this.states = {...states};
```
QingWei-Li: We can use the [`merge` function](https://github.com/ElemeFE/element/blob/dev/src/utils/merge.js).

```js
this.states = merge({}, states)
```

QingWei-Li: We can use the [`merge` function](https://github.com/ElemeFE/element/blob/dev/src/utils/merge.js).

```js
this.states = merge({}, states)
```

kolesoffac: Ok, i need edit it or you?

QingWei-Li: You do it if you have time

kolesoffac: Good, i will do it

PR #9053, opened 2017-12-28
Update fr.js

kompass:
In french, like in english, we don't say "Année" after the year number.


mathieutu:
@kompass FYI I just PRed that: https://github.com/ElemeFE/element/pull/9056

kompass:
@mathieutu Thank you

PR #9188, opened 2018-01-09
pr #8940

kolesoffac:
Update code for #8940

### DIFF
--- a/packages/table/src/table-store.js
+++ b/packages/table/src/table-store.js
@@ -90,6 +90,9 @@ const TableStore = function(table, initialState = {}) {
     leafColumns: [],
     fixedLeafColumns: [],
     rightFixedLeafColumns: [],
+    leafColumnsLength: 0,
+    fixedLeafColumnsLength: 0,
+    rightFixedLeafColumnsLength: 0,
     isComplex: false,
     _data: null,
     filteredData: null,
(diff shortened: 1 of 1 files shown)


Leopoldthecoder:
Hi, thanks for contributing!

This fix is valid, but merging a big object may lead to some performance issues. 

I suppose we have two alternatives: we can replace

https://github.com/ElemeFE/element/blob/0d952b79174520957e86dc5f4bf6cccd96a58ca0/packages/table/src/table-store.js#L349-L351

with

```JS
Vue.set(states, 'leafColumnsLength', leafColumns.length);
Vue.set(states, 'fixedLeafColumnsLength', fixedLeafColumns.length);
Vue.set(states, 'rightFixedLeafColumnsLength', rightFixedLeafColumns.length);
```

Or simply declare these 3 keys when initiating `this.states`:
https://github.com/ElemeFE/element/blob/0d952b79174520957e86dc5f4bf6cccd96a58ca0/packages/table/src/table-store.js#L83-L109

I prefer the second approach, what do you think?

kolesoffac:
I think, that first with Vue.set.

liril-net:
I do believe that the second one would be better. Cause state data flow can be controlled in this way.

Leopoldthecoder:
@kolesoffac I took the liberty of updating this PR to my proposal, hope you don't mind. Again, thanks for your contribution.

PR #9310, opened 2018-01-17
DatePicker: fix incorrect interaction on range change when defaultTime provided

nighca:
Fix interaction issue introduced in https://github.com/ElemeFE/element/pull/9094#issuecomment-357947605

### DIFF
--- a/packages/date-picker/src/panel/date-range.vue
+++ b/packages/date-picker/src/panel/date-range.vue
@@ -426,9 +426,8 @@
       },
 
       handleChangeRange(val) {
-        const defaultTime = this.defaultTime || [];
-        this.minDate = modifyWithGivenTime(val.minDate, defaultTime[0]);
-        this.maxDate = modifyWithGivenTime(val.maxDate, defaultTime[1]);
+        this.minDate = val.minDate;
+        this.maxDate = val.maxDate;
         this.rangeState = val.rangeState;
       },
 
--- a/test/unit/specs/date-picker.spec.js
+++ b/test/unit/specs/date-picker.spec.js
@@ -1272,20 +1272,23 @@ describe('DatePicker', () => {
           triggerEvent(leftCell, 'click', true);
           setTimeout(_ => {
             triggerEvent(rightCell, 'mousemove', true);
-            triggerEvent(rightCell, 'click', true);
-
             setTimeout(_ => {
-              const {
-                minDate,
-                maxDate
-              } = vmWithDefaultTime.picker;
-              expect(minDate.getHours()).to.be.equal(11);
-              expect(minDate.getMinutes()).to.be.equal(59);
-              expect(minDate.getSeconds()).to.be.equal(59);
-              expect(maxDate.getHours()).to.be.equal(0);
-              expect(maxDate.getMinutes()).to.be.equal(0);
-              expect(maxDate.getSeconds()).to.be.equal(0);
-              done();
+              expect(rightCell.classList.contains('in-range')).to.be.true;
+
+              triggerEvent(rightCell, 'click', true);
+              setTimeout(_ => {
+                const {
+                  minDate,
+                  maxDate
+                } = vmWithDefaultTime.picker;
+                expect(minDate.getHours()).to.be.equal(11);
+                expect(minDate.getMinutes()).to.be.equal(59);
+                expect(minDate.getSeconds()).to.be.equal(59);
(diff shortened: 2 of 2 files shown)


test/unit/specs/date-picker.spec.js
```diff
@@ -1272,20 +1272,23 @@ describe('DatePicker', () => {
...
-                maxDate
-              } = vmWithDefaultTime.picker;
-              expect(minDate.getHours()).to.be.equal(11);
-              expect(minDate.getMinutes()).to.be.equal(59);
-              expect(minDate.getSeconds()).to.be.equal(59);
-              expect(maxDate.getHours()).to.be.equal(0);
-              expect(maxDate.getMinutes()).to.be.equal(0);
-              expect(maxDate.getSeconds()).to.be.equal(0);
-              done();
+              expect(rightCell.className.split(' ').indexOf('in-range') >= 0).to.be.true;
```
wacky6: How about using `classList` ?

packages/date-picker/src/panel/date-range.vue
```diff
@@ -426,9 +426,8 @@
       },
 
       handleChangeRange(val) {
-        const defaultTime = this.defaultTime || [];
-        this.minDate = modifyWithGivenTime(val.minDate, defaultTime[0]);
-        this.maxDate = modifyWithGivenTime(val.maxDate, defaultTime[1]);
+        this.minDate = val.minDate;
```
wacky6: I think it's better to "trim" time before passing {min, max}Date to `date-table`.

test/unit/specs/date-picker.spec.js
nighca: I'm not sure about the browser support requirement of element. But it is safer to use `className` and almost the same complex.

packages/date-picker/src/panel/date-range.vue
nighca: I cancelled my previous changes here to fix this issue (now it's just the same behavior as what it used to be). If we want to add `trim` behavior before passing `minDate` / `maxDate` to `date-table`, maybe that should be another issue?

test/unit/specs/date-picker.spec.js
wacky6: Tests are executed in Chrome. so I don't think this will be an issue.

packages/date-picker/src/panel/date-range.vue
wacky6: Actually, I mean resetting the time to 00:00:00 before passing minDate/maxDate to date-table.

date-table checks date equality in a naive way, so it can not handle non 00:00:00 time. But the rest of date-range panel should know about the selected time.

Like this:
```
```

test/unit/specs/date-picker.spec.js
nighca: Maybe possible future e2e tests in different browsers? If browser supporting has never become an issue when writing test cases in this project, I'll use `classList.contains` instead.

packages/date-picker/src/panel/date-range.vue
nighca: That makes sense. Thx.

nighca: Seems it is a little complex to realize this.

1. I have to reset to `12:00:00` to make it work (I'm still looking into why), which will change the default behavior on pick.

2. There is some "reference equal" issue, which means, when `resetTime` returns a new `Date` object (even with the same time value), it breaks.

I'll spend time on it later. If you have any ideas about the above problems, plz hint me.

test/unit/specs/date-picker.spec.js
wacky6: You can safely use classList. 2.0 requires ie >10 which supports it.  see https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

packages/date-picker/src/panel/date-range.vue
wacky6: Interesting. Might have something to do with legacy codes. I suggest you keep current solution, it works. :)

test/unit/specs/date-picker.spec.js
nighca: great.

packages/date-picker/src/panel/date-range.vue
nighca: Agree that.

PR #9320, opened 2018-01-17
Slider: Touch events added to support mobile devices

vladislav-sevostyanov:


packages/slider/src/button.vue
```diff
@@ -130,7 +131,9 @@
         event.preventDefault();
         this.onDragStart(event);
         window.addEventListener('mousemove', this.onDragging);
+        window.addEventListener('touchmove', this.onDragging);
         window.addEventListener('mouseup', this.onDragEnd);
+        window.addEventListener('touchend', this.onDragEnd);
```
syn-zeta: The listeners should be removed `onDragEnd` as well.

packages/slider/src/button.vue
```diff
@@ -189,6 +200,7 @@
           }, 0);
           window.removeEventListener('mousemove', this.onDragging);
           window.removeEventListener('mouseup', this.onDragEnd);
+          window.removeEventListener('touchend', this.onDragEnd);
```
syn-zeta: Please remove `touchmove` as well, added on line 134 above.

vladislav-sevostyanov:
It's done, comrade.

wacky6:
So element is now accepting mobile optimization PRs?

syn-zeta:
Is touch support considered mobile only? Most manufacturers offer at least a couple laptop/all-in-one touch options.

wacky6:
Never had an all-in-one, not sure whether browsers in all-in-one devices emit touch events.
If they don't, for me, touch events belong to mobile / tablet optimization.

syn-zeta:
Yes, laptops/desktops with touchscreen emit touch events.

wacky6:
Then sounds like element-ui is going to be touch-friendly and accidentally become usable in touch-only environments. :)

Leopoldthecoder:
My concern is that we don't want the library to be bloated with mobile-devide-only code. But this PR has only 13 lines of additions, so I merged it.

PR #9329, opened 2018-01-18
[Select] [Option] convert string like number to convert to number

rayqian:
Add a `isNumberic` function check to ensure if option's value is a string like number, then convert it to number, then compare to the select's value.

Because `PHP` always give us a string number, in this way, we can ignore the type of the value.


Leopoldthecoder:
We will not support this. You should do the conversion in the parent component.

PR #9363, opened 2018-01-19
Table: Add indeterminate & disabled state to multiple selection header checkbox

### LINKED ISSUE #9508: [Bug Report] Multi-select table's checkbox does not show indeterminate state as Transfer component
### Element UI version
2.0.11

### OS/Browsers version
OS X 10.11 / Chrome 62

### Vue version
2.5.13

### Reproduction Link
[https://jsfiddle.net/3jn2ftsz/](https://jsfiddle.net/3jn2ftsz/)

### Steps to reproduce
- In a multi-select table with more than one row, select one of the rows by clicking on the checkbox to the left.
- Notice that the select-all checkbox in the header does not indicate that some items in the table are checked.
- For [Transfer](http://element.eleme.io/#/en-US/component/transfer), notice that the checkbox in the header indicates with an indeterminate state that some items in the list below are checked.

### What is Expected?
Consistent UX of select-all checkbox in multi-select components like Table and Transfer.

Multi-select Table's select-all checkbox does not have an indeterminate state when only some of the rows are checked.

syn-zeta:
This indeterminate state when only some rows are selected will make the component behavior consistent with that of the Transfer component.

Fixed #9508

### DIFF
--- a/packages/table/src/table-column.js
+++ b/packages/table/src/table-column.js
@@ -34,7 +34,6 @@ const forced = {
   selection: {
     renderHeader: function(h, { store }) {
       return <el-checkbox
-        disabled={ !store.states.filteredData.length }
         indeterminate={ store.states.selection.length > 0 && !this.isAllSelected }
         nativeOn-click={ this.toggleAllSelection }
         value={ this.isAllSelected } />;
--- a/packages/table/src/table-column.js
+++ b/packages/table/src/table-column.js
@@ -32,8 +32,10 @@ const defaults = {
 
 const forced = {
   selection: {
-    renderHeader: function(h) {
+    renderHeader: function(h, { store }) {
       return <el-checkbox
+        disabled={ !store.states.filteredData.length }
+        indeterminate={ store.states.selection.length > 0 && !this.isAllSelected }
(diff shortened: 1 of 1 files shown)


syn-zeta:
The disabled checkbox still gets checked on click as the native event is handled instead of Checkbox's change event. What's the best place to address that?

syn-zeta:
@QingWei-Li @Leopoldthecoder Requesting review for this small diff.

syn-zeta:
Fixes #9508

Leopoldthecoder:
Maybe for now we can ignore the disabled state. The header checkbox is only disabled when all rows are uncheckable, which does not happen quite often. Plus, computing if all rows are uncheckable may be a performance issue.

So you can remove line 37.

syn-zeta:
@Leopoldthecoder Removed.

Although, the condition that existed to disable the checkbox was when no rows were visible in the table. Is that not a case when we would like to disable it?

Leopoldthecoder:
@syn-zeta OK, I'll take care of that.

JChermy:
请问有提供参数支持去掉table中的全选按钮吗？不只是在数据为空时disable

PR #9501, opened 2018-01-28
DateTimePicker: fix invalid month value bug for Firefox 37

c0hb1rd:
NeoKylin OS default browser is Firefox 37 and the Date function required two-digit month value.


c0hb1rd:
@Leopoldthecoder Thx . Indeed your solution is  better : )

PR #9543, opened 2018-01-30
Table: add disabled state to multiple selection header checkbox

Leopoldthecoder:

### DIFF
--- a/packages/table/src/table-column.js
+++ b/packages/table/src/table-column.js
@@ -34,6 +34,7 @@ const forced = {
   selection: {
     renderHeader: function(h, { store }) {
       return <el-checkbox
+        disabled={ store.states.data && store.states.data.length === 0 }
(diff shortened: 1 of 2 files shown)


packages/table/src/table-column.js
```diff
@@ -34,6 +34,7 @@ const forced = {
   selection: {
     renderHeader: function(h, { store }) {
       return <el-checkbox
+        disabled={ store.states.data && store.states.data.length === 0 }
```
syn-zeta: @Leopoldthecoder 
Should `filteredData` be referenced instead of `data` in case there are some column filters applied and the table does not have any visible rows but still has some data?

Leopoldthecoder: `data` is already filtered. The difference between `data` and `filteredData` is that `data` is sorted (if the table can be sorted). The original data is stored in `_data`.

syn-zeta: Ah, okay.

PR #9607, opened 2018-02-01
Fix: all docs about port

lixinGiting:
In setting port = 8085, but in the documents explain visit localhost:8086.


Leopoldthecoder:
This doc is for when you start your own project using Element, not for starting Element itself.

That being said, since now we encourage people to start their project using vue-cli, I'll remove this line.

PR #9640, opened 2018-02-02
Upload: allow custom file icon

matjaz:
Allow custom file icon class. Class can be updated on `change` or any other event.


Leopoldthecoder:
Sorry, we may not support this.

matjaz:
@Leopoldthecoder  Can you elaborate? And what would be the recommended workaround?

Leopoldthecoder:
You can use the default slot to write your own file list.

PR #9757, opened 2018-02-09
Popper: Fix memory leak on route change (#7715)

### LINKED ISSUE #7715: [Bug Report] Memory leak on route change with a visible tooltip
### Element UI version
1.4.8

### OS/Browsers version
Mac OS X 10.11.6 / Chrome 61.0.3163.100 (Official Build) (64-bit)

### Vue version
2.5.2

### Reproduction Link
[https://jsfiddle.net/kydfpodc/1/show/](https://jsfiddle.net/kydfpodc/1/show/)

### Steps to reproduce
- Click on "go to Foo" to load the table initially
- Decrease viewport width until the address column in the table has ellipsis, and shows tooltips on hover
- Click on one of the rows in the address column while the tooltip is visible to be redirected to "Bar"
- Click on "go to Foo" to view the table again

### What is Expected?
Node count after step 4 to be the same as the node count after step 1.

Steadily increasing node count as the detached tables remain in memory due to references held in tooltips.

image

image

syn-zeta:
#### Fixes #7715 by addressing the two following issues:
- When the route changed on click, the `doDestroy` method of vue-popper was returning right away without actually destroying the popper due to `showPopper` being true. `forceDestroy` parameter is added to handle this case. An alternative fix could be to set `showPopper` to `false` in `beforeDestroy` before calling `doDestroy`, but the parameter approach was chosen as it can be reused.
- After the above fix, the `destroy` method of popper.js was being successfully called, which in turn called `_removeEventListeners`. The `target` that this method was removing the `scroll` event listener from was not the same element to which the listener was added (this is due to `getScrollParent` not being able to find the correct element as some of the top level nodes had already been destroyed). This caused the event listener (bound to `this`) to leak, and consequently the whole associated DOM and VDOM tree. The fix is to replace the logic to find the target in `_removeEventListeners` and use the stored target to remove the listener from.

### DIFF
--- a/src/utils/popper.js
+++ b/src/utils/popper.js
@@ -240,7 +240,6 @@
         if (typeof this.state.updateCallback === 'function') {
             this.state.updateCallback(data);
         }
-
     };
 
     /**
@@ -438,7 +437,6 @@
         popperOffsets.width   = popperRect.width;
         popperOffsets.height  = popperRect.height;
 
-
         return {
             popper: popperOffsets,
             reference: referenceOffsets
@@ -464,6 +462,7 @@
                 target = root;
             }
             target.addEventListener('scroll', this.state.updateBound);
+            this.state.scrollTarget = target;
         }
     };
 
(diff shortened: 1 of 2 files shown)


syn-zeta:
#### Second issue:
popper_out

#### With the fix:
fix

PR #9863, opened 2018-02-23
[Table]: Add information under table header about how many rows were selected as well as provide options for bulk actions

smilee:
Please review and let me know what you think!
Oh and please fix any problems that may arise with responsive layouts. Thank you! 🙇‍♀️


smilee:
I am working on a better solution.

PR #9904, opened 2018-02-27
feat(tansfer): target value added to top, fixed #9882

### LINKED ISSUE #9882: [Feature Request] Disable el-transfer auto-sort
### Existing Component
Yes

### Component Name
Transfer

### Description
When i move an item between lists the list getting sorted automatically actually i don't want that feature , i really want that as an optional , in my scenario when a new item added to the list it should be in the last on the list , but due to auto sort the whole list rearranges , can you please able to make auto sort as optional , please add a parameter to disable this feature that would be great.
Thanks

QingWei-Li:

Auto format...

Only added this code
```js
  computed: {
    dataObj() {
      const key = this.props.key;
      return this.data.reduce((o, cur) => (o[cur[key]] = cur) && o, {});
    },
 
    sourceData() {
      return this.data.filter(
        item => this.value.indexOf(item[this.props.key]) === -1
      );
    },
 
    targetData() {
      return this.value.map(key => this.dataObj[key]);
    },
```

### DIFF
--- a/packages/transfer/src/main.vue
+++ b/packages/transfer/src/main.vue
@@ -42,158 +42,165 @@
 </template>
 
 <script>
-  import ElButton from 'element-ui/packages/button';
-  import Emitter from 'element-ui/src/mixins/emitter';
-  import Locale from 'element-ui/src/mixins/locale';
-  import TransferPanel from './transfer-panel.vue';
-  import Migrating from 'element-ui/src/mixins/migrating';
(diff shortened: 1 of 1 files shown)


Leopoldthecoder:
反了，刚移过来的项应该在最下面。

Leopoldthecoder:
我提了一个 #9960，开了个新属性，默认还是保持现有行为吧。

PR #9924, opened 2018-02-28
Multi-select Table: Add selectOnIndeterminate prop to control master checkbox behavior

syn-zeta:
Makes the behavior of the select-all checkbox same as GMail -- to deselect in "indeterminate" state of checkbox:

table

### DIFF
--- a/packages/table/src/table.vue
+++ b/packages/table/src/table.vue
@@ -306,7 +306,12 @@
 
       tooltipEffect: String,
 
-      spanMethod: Function
+      spanMethod: Function,
+
+      selectOnIndeterminate: {
+        type: Boolean,
+        default: true
+      }
     },
 
     components: {
@@ -618,7 +623,8 @@
     data() {
       const store = new TableStore(this, {
         rowKey: this.rowKey,
-        defaultExpandAll: this.defaultExpandAll
+        defaultExpandAll: this.defaultExpandAll,
+        selectOnIndeterminate: this.selectOnIndeterminate
       });
       const layout = new TableLayout({
         store,
--- a/examples/docs/en-US/table.md
+++ b/examples/docs/en-US/table.md
@@ -1946,7 +1946,7 @@ You can customize row index in `type=index` columns.
 
 ### Table Attributes
 | Attribute      | Description          | Type      | Accepted Values       | Default  |
-|---------- |-------------- |---------- |--------------------------------  |-------- |
+|----------------|----------------------|-----------|-----------------------|----------|
 | data | Table data | array | — | — |
 | height | Table's height. By default it has an `auto` height. If its value is a number, the height is measured in pixels; if its value is a string, the value will be assigned to element's style.height, the height is affected by external styles | string/number | — | — |
 | max-height | Table's max-height. The height of the table starts from `auto` until it reaches the maxHeight limit. The `maxHeight` is measured in pixels, same as `height` | string/number | — | — |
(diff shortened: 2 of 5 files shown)


wacky6:
Failed at irrelevant CI tests, try re-run

syn-zeta:
@wacky6 Passed.

syn-zeta:
This behavior can also be controlled by a prop, and I can update the PR if we'd like that.

QingWei-Li:
@syn-zeta Yes please.

syn-zeta:
@QingWei-Li Added `select-on-indeterminate` prop (couldn't think of a better name).

syn-zeta:
@QingWei-Li Unrelated to this PR, but I cannot find any better place to reach the maintainers regarding this question that I'd posted on Gitter some time ago:

untitled

syn-zeta:
@QingWei-Li Leo seems busy. Is there anyone else who'd be willing to review this PR? It'd be nice to have it in the next release. Also, would the Element team be willing to share some specifics of their icon font workflow (ref. above comment)?

packages/table/src/table.vue
```diff
@@ -306,7 +306,9 @@
 
       tooltipEffect: String,
 
-      spanMethod: Function
+      spanMethod: Function,
+
+      selectOnIndeterminate: Boolean
```
Leopoldthecoder: Hi, sorry for the delay. I think this prop should default to `true` so that the default behavior is not changed. Otherwise it's kind of a breaking change.

test/unit/specs/table.spec.js
```diff
@@ -815,10 +815,10 @@ describe('Table', () => {
     });
 
     describe('type', () => {
-      const createTable = function(type) {
+      const createTable = function(type, tableProps) {
         return createVue({
           template: `
-            <el-table :data="testData" @selection-change="change">
+            <el-table :data="testData" @selection-change="change" ${tableProps || ''}>
```
Leopoldthecoder: Is these additions necessary?

Leopoldthecoder:
@syn-zeta I talked to our designers about the icon font. When they drew the svg source files they had a 16x16 border for every icon, which means every svg file has a dimension of 16x16. Then we converted the svg files into icon fonts.

packages/table/src/table.vue
syn-zeta: No problem, thanks for the review.

It would be a breaking change, yes, but arguably a better default behavior? Regardless, I've updated it to default to `true`.

test/unit/specs/table.spec.js
syn-zeta: I was thinking of adding unit tests to cover this prop, but wasn't sure if they should be added to `describe('attributes'` or `describe('type'`.

I've removed the changes for now, but can add them in a new PR if required.

syn-zeta:
>When they drew the svg source files they had a 16x16 border for every icon
Thanks for following-up on this. Would you be able to ask them what is meant by having a "border" for every icon? What tool/what setting would allow setting a "border"? If you and they don't mind, I can also contact them directly for some directions on this.

Leopoldthecoder:
@syn-zeta Could you tell me what your email is? They gave me the icon template source file.

syn-zeta:
@Leopoldthecoder Oh, great! My email is `synthoidzeta [at] gmail [dot] com`. Thank you very much!

PR #10014, opened 2018-03-05
Add emit destroy popover

kolesoffac:
It need when run action after end animation(hide).

Thanks

### DIFF
--- a/packages/popover/src/main.vue
+++ b/packages/popover/src/main.vue
@@ -122,7 +122,7 @@ export default {
   methods: {
     doPrivateDestroy() {
       this.doDestroy();
-      this.$emit("distroyed");
+      this.$emit('distroyed');
     },
     doToggle() {
       this.showPopper = !this.showPopper;
--- a/packages/popover/src/main.vue
+++ b/packages/popover/src/main.vue
@@ -1,6 +1,6 @@
 <template>
   <span>
-    <transition :name="transition" @after-leave="doDestroy">
+    <transition :name="transition" @after-leave="doPrivateDestroy">
       <div
         class="el-popover el-popper"
         :class="[popperClass, content && 'el-popover--plain']"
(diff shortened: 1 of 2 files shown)


packages/popover/src/main.vue
```diff
@@ -120,6 +120,10 @@ export default {
   },
 
   methods: {
+    doPrivateDestroy() {
+      this.doDestroy();
+      this.$emit('distroyed');
```
syn-zeta: `destroyed`. The documentation will also need to be updated.

Can you rely on the existing `hide` event instead (emitted on line 64 above)? Why is this needed?

kolesoffac: *Hide* triggered immediate, when transition don't finish yet! But sometimes need call action after full hide *popover*.

Leopoldthecoder:
@kolesoffac Have you tried `hide` event?

packages/popover/src/main.vue
kolesoffac: >>The documentation will also need to be updated.

But where?

Leopoldthecoder: @kolesoffac It's [here](https://github.com/ElemeFE/element/blob/dev/examples/docs/en-US/popover.md).

Leopoldthecoder:
@kolesoffac If that's the case, I'm afraid its counterpart in the entering transition is also needed. Perhaps we can call them `after-enter` and `after-leave`.

If you're interested, you can update this PR.

kolesoffac:
Are you interest and enter, and leave of transition?

syn-zeta:
@kolesoffac Yes, it'd be nice to see both `after-enter` and `after-leave` events, in that case. There's also documentation in three languages, so you can also just copy-paste English description in other ones, and someone else will translate them (they'll act as placeholders until then, and the docs won't get out-of-sync).

Leopoldthecoder:
I'll add `after-enter` and `after-leave` events right away. Closing this PR.

Leopoldthecoder:
#10047

Issue #10008, opened 2018-03-05
[Bug Report] Select element can't be filtered if down key isn't pressed beforehand


Terrijoo:
### Element UI version
2.2.1

### OS/Browsers version
Version 64.0.3282.186 (64-Bit)

### Vue version
2.5.13

### Reproduction Link
[https://jsfiddle.net/Terrijoo/d7L01ufv/](https://jsfiddle.net/Terrijoo/d7L01ufv/)

### Steps to reproduce
Click into the first Input field, press "Tab" and try to type.

### What is Expected?
The list pops up and it starts filtering.

### What is actually happening?
You can't type in it at all, you need to explicitly press down beforehand to let the list pop up.

wacky6:
Seems to be a regression in 2.2.1.

wacky6:
Related to PR https://github.com/ElemeFE/element/pull/9857 and https://github.com/ElemeFE/element/pull/9894

@Seebiscuit  @openks Could you guys take a look? Expected behavior is: https://jsfiddle.net/d7L01ufv/5/

Seebiscuit:
I believe the behavior stems from this deletion

```js
       handleFocus(event) {
         this.visible = true; // removed by #9894
         this.$emit('focus', event);
       },
```

Setting true to `this.visible` reveals the popper. Perhaps @openks or @leopoldthecoder can shed more light on the issue, and explain why `this.visible` was removed in that PR.

openks:
according to #9821 the select or the input in select focus show not open the popper,
only select clicked or the input in select focused and the enter key pressed the popper show

Terrijoo:
@openks Is there at least a possibility to make it optional, like an extra attribute? It can be very a very tedious work to fill a form with a lot of fields, if you tab through them and have to first press the down key everytime.

openks:
The original select when you tab  through is also only focused and you should press the enter key to  show the options ,so i think it's ok 
@Leopoldthecoder i don't know is it ok to add a extra attribute to control this situation? if it's ok i think i can do it .

Leopoldthecoder:
This actually is intended in 2.2.1. But since you brought it up, we may now discuss if it is reasonable.

Before 2.2.1, the dropdown shows when you focus the Select. This makes navigating or filling the forms easier, but causes #9821. Plus, the native `<select>` does not show its options on focus.

In 2.2.1, Select behaves just the opposite. So the pros and cons are turned.

[...]

PR #10042, opened 2018-03-07
Allows user to choose if select focus triggers menu

Seebiscuit:
Addresses #10008. Allows user to choose if tabbing to select element toggles menu. 

Added `automatic-dropdown` to input elements of `select`. When component receives focus, `handleFocus` is invoked. Inside, the prop `automaticDropdown`, bound to the `automatic-dropdown` attribute, is tested. If the attribute is selected `this.visible` is set to `true`. The test is on a timeout, because on mousedown the input element will immediately emite a `focus` event that bubbles to the `select` parent. So, on click, `handleFocus` fires on the mousedown portion of that click and will always fire before the `click` event that is bound to `toggleMenu`. The timeout ensures that the `automaticDropdown` prop is tested _after_ `toggleMenu` handles opening the menu.  

If PR accepted, I can also update the docs to include the new attribute.


Seebiscuit:
Not sure why the last 7 commits where included in this PR. Only the last commit should've been part of this PR.

Leopoldthecoder:
It seems you didn't pull the latest code from the upstream `dev` branch. Please do that and then commit your changes.

Seebiscuit:
No, I pulled the code. Do I need to rebase?

Leopoldthecoder:
Yes, `git pull [upstream] dev --rebase`.

Seebiscuit:
Doing pull -r did everything's up to date... 🙁 Resolved the conflicts and merged commits

packages/select/src/select.vue
```diff
@@ -48,6 +48,7 @@
         :class="[selectSize ? `is-${ selectSize }` : '']"
         :disabled="selectDisabled"
         :autocomplete="autoComplete"
+        :automatic-dropdown="automaticDropdown"
```
Leopoldthecoder: What the use of this prop? `<input>` and `<el-input>` don't have such API.

Seebiscuit:
That's the bais of the PR. In #10008 we discussed the idea that select menus would not open on focus, but that the user could override that behavior with an attribute. If `automatic-dropdown` is set to true, then the menu will open on focus.

Leopoldthecoder:
Hi @Seebiscuit , it seems the newly added test cases have failed, could you please take a look?

Seebiscuit:
I've been trying to investigate. I'm having a really hard time getting the debugger to stop on that code. I can debug if I remove `--single-run` as a param to karma, but the tests don't compete. On the other hand if I use `--single-run` and increase the timeout so I can break the debugger, I still get "no response for over 10000 ms" messages.

How do you recommend to debug tests?

wacky6:
@Seebiscuit I usually use developer console to run test codes manually in browser.

Seebiscuit:
How would you run them manually? Without karma?

wacky6:
* `yarn dev` (opens dev server)
* goto `http://localhost:8085/`, goto component's doc page 
* may need to modify doc's source code to reflect the test in question
* read the test code, copy & paste into console, modify variable references
* inspect the component state

Seebiscuit:
Ahhh. I guess we can try that. Good call!

Terrijoo:
Don't wanna sound pushy, but any news on this?

test/unit/specs/select.spec.js
```diff
@@ -701,6 +701,54 @@ describe('Select', () => {
...
+        </div>
+      `,
+      data() {
+        return {
+          value: ''
+        };
+      }
+    }, true);
+
+    vm.$refs.select.$refs.reference.$el.focus();
```
wacky6: reference to input should be `vm.$refs.select.$refs.reference.$refs.input`; $el references to el-input's wrapping `div` which does not focus.

test/unit/specs/select.spec.js
```diff
@@ -701,6 +701,54 @@ describe('Select', () => {
...
+
+    setTimeout(() => {
+      expect(vm.visible).to.be.false;
+      done();
+    }, 310);
```
wacky6: timeout is generally `10`

test/unit/specs/select.spec.js
```diff
@@ -701,6 +701,54 @@ describe('Select', () => {
...
+        return {
+          value: ''
+        };
+      }
+    }, true);
+
+    vm.$refs.select.$refs.reference.$el.focus();
+
+    setTimeout(() => {
+      expect(vm.visible).to.be.false;
```
wacky6: reference to el-select should be `vm.$refs.select`

packages/select/src/select.vue
```diff
@@ -528,6 +529,7 @@
 
       handleFocus(event) {
         if (!this.softFocus) {
+          setTimeout(() => { if (this.automaticDropdown) this.visible = true; }, 300);
```
wacky6: Why 300ms timeout? Why not set `visible` straightway?

wacky6:
@Seebiscuit I've checked your code, see comments for fixes.

test/unit/specs/select.spec.js
Seebiscuit: Ok. Will implmenet and run the test again.

packages/select/src/select.vue
Seebiscuit: The `automaticDropdown` check is on a timeout, because on `mousedown` the **`input`** element will immediately emit an **`onfocus` event** that bubbles to the **`select`** component and fires `handleFocus`. The  `automaticDropdown` check is inside `handleFocus` and therefore fires on focus which happens either `onmousedown` (which is when `onfocus` is triggered) or after a tab operation lands on the element. On the other hand, `toggleMenu`, which checks if `!this.visible`,  fires ***after*** `onclick` (and therefore always after `handleFocus`). 

If  the `automaticDropdown` check fires too soon, then `this.visible === true` by the time `toggleMenu` hits, and the menu will always remain closed. By setting a timeout of 300ms the check waits long enough to fire *after* `onclick`--*in the majority of cases*; but not too long that it feels totally unresponsive.

I originally wanted to make the `automaticDropdown` check more intelligent by checking to see if `handleFocus` was invoked by an `onmousedown` event, but that's not possible because`handleFocus` is bound to a `onfocus` event which fires as a result of `onmousedown`

This discussion also answers your questions regarding the timeout on the test. To properly test this feature we have to wait the 300 ms for it to activate.

wacky6: Personally, I'm not a big fan of timeout. Timing issues is not really something we want in our code. It's hard to debug if problem arises, and may lead to flaky tests.

My idea is to add a flag to record whether `automaticDropdown` just set `visible` to `true`; Check that flag in `toggleMenu`; Then clear the flag when `visible` state changes. This hack appears to be working on simple cases.

@Leopoldthecoder Any better ideas?

Leopoldthecoder: I agree with @wacky6 .

- After setting `visible` to true in this `if` block, also set a flag to `true`
- In `toggleMenu`, check this flag, and do the actual toggling only when the flag is `false`
- Reset this flag to `false` in the `visible` watcher

Also, please remember to reset `softFocus` to `false` in `handleBlur`.

Seebiscuit: Ok. Will do. Will update soon.

Seebiscuit:
Hey @wacky6 @Terrijoo and @Leopoldthecoder any update on when this will be merged?

packages/select/src/select.vue
Leopoldthecoder: @Seebiscuit Hi, please update this PR based on discussions here.

Seebiscuit: Yea, I'm sorry, I overlooked @wacky6 comment. The update was made in the last commit.

wacky6:
* When use `remote-method` or `allow-create` without `automatic-dropdown`, tab into input and type, results in strange behavior. 
* When use `filterable` without `automatic-dropdown`, tab into input and type, dropdown is not automatically shown.

Need to update documentation to reflect this (or imply automatic-dropdown for these situations).

Other than these two edge cases, basic functionality seems ok. Should be good enough for merge.

Seebiscuit:
Ok if I add documentation to this PR or should I pull a new request? I'm not going to address @wacky6 issues with `remote-method`/`filterable`. Rather, just add `automatic-dropdown` to the `props` table.

wacky6:
We will do the docs. Thanks for contrib

Seebiscuit:
You got it. This was my entry in the `props` table, in case you find it useful:
```
| automatic-dropdown | whether a select menu should appear when the element gains focus from a tab operation | boolean | — | false |
```

Issue #10173, opened 2018-03-14
[Feature Request] Bugs with format for el-date-picker (datetime)


remizovvv:
### Existing Component
Yes

### Component Name
DatePicker

### Description
We can set format property for el-date-picker. But if type=datetime, formats of "date" input and timepicker in share picker panel are not calculated correctly. You can see this bug in this fiddle:

https://jsfiddle.net/remizovvv/L4h9tkyc/3/

"Parent" input - view date-time correct, but "date" input - displays date and some part of time format (hours and AM/PM). "Time" input ignores setted format.
I try added some more correctly extracting time format and date format from format property. And added two new properties: time-format and date-format, if users will be want manual set  this formats.

Leopoldthecoder:
Merged #10174.

Issue #9878, opened 2018-02-24
[Feature Request] DatePicker Design Improvement


wacky6:
### Description
`{Date, Time, DateTime}Picker`, including `{Year, Month, Week, *Range}` derivatives, **are bundled together**. This makes the output size unnecessarily large.

In fact, it is one of the largest components in element-ui.

### Reasons:

* If one only uses DatePicker, other panels are still bundled. (wasted bytes)
* Improve code readability / maintainability (reduce 'if's regarding `type`, `isRanged`, etc)
* Encourage type / panel specific behaviors without having to worry about others
* Move `picker-options` to component tag, improve readability?
* Reuse / share basic components ({date, month, year} table)
* Split documentation, so developers are less likely to get confused about attributes

Therefore, I make this proposal to bundle/write these components separately. It could follow ant.d's style: https://ant.design/components/date-picker/ , for example: 

```jsx
<el-date-picker />
<el-date-time-picker />
<el-week-picker />
<el-time-picker />
...
```

Will be a **breaking change**, might be considered for 3.0 milestone.

### Additional Notes
* Expose underlying {date, month, year} table, so developers can build custom components https://github.com/ElemeFE/element/issues/12779. (eg: inline date-picker or simple calendar)
* Offer a common interface, so Time{Select, Picker} can be used interchangeably, see: https://github.com/ElemeFE/element/issues/8979, https://github.com/ElemeFE/element/issues/8912, https://github.com/ElemeFE/element/issues/11818, https://github.com/ElemeFE/element/issues/11594, https://github.com/ElemeFE/element/issues/13766 (i.e `value`'s format, input/change event emission behavior), 
* Support for week format (i.e. ISO-8601). May need to change underlying parser. Current parsing / formatting method is inappropriate (not modular). https://github.com/ElemeFE/element/issues/13746 https://github.com/ElemeFE/element/pull/13754
* Reactive panel type is possible. But I suspect for most use cases, panel type is fixed. If panel type does needs to be changed, use v-if to re-render the entire component is more appropriate: https://github.com/ElemeFE/element/issues/12119
* Expose underlying fecha parser / formatter
* Might be interested in https://github.com/linusborg/portal-vue to render panels elegantly
* Support custom format / parse function: https://github.com/ElemeFE/element/issues/11353 https://github.com/ElemeFE/element/issues/14612
* Better way to set disabled date and disabled time (group of ranges maybe?), try to improve performance?
[...]

PR #10174, opened 2018-03-14
DatePicker: added correct set time and date formats

remizovvv:
DatePicker: added correct extracting time format and date format from `format` property (datatimepicker). And added special property `date-format` and `time-format`, for manual set

https://github.com/ElemeFE/element/issues/10173

### DIFF
--- a/packages/date-picker/src/panel/date.vue
+++ b/packages/date-picker/src/panel/date.vue
@@ -153,7 +153,9 @@
     nextYear,
     prevMonth,
     nextMonth,
-    changeYearMonthAndClampDate
+    changeYearMonthAndClampDate,
+    extractDateFormat,
+    extractTimeFormat
   } from '../util';
   import Clickoutside from 'element-ui/src/utils/clickoutside';
   import Locale from 'element-ui/src/mixins/locale';
@@ -215,7 +217,6 @@
         const value = value => {this.$refs.timepicker.value = value;};
         const date = date => {this.$refs.timepicker.date = date;};
 
-        this.$watch('format', format);
         this.$watch('value', value);
         this.$watch('date', date);
 
@@ -523,8 +524,8 @@
       },
 
       timeFormat() {
-        if (this.format && this.format.indexOf('ss') === -1) {
-          return 'HH:mm';
+        if (this.format) {
+          return extractTimeFormat(this.format);
         } else {
           return 'HH:mm:ss';
         }
@@ -532,7 +533,7 @@
 
       dateFormat() {
         if (this.format) {
-          return this.format.replace('HH', '').replace(/[^a-zA-Z]*mm/, '').replace(/[^a-zA-Z]*ss/, '').trim();
+          return extractDateFormat(this.format);
         } else {
           return 'yyyy-MM-dd';
         }
--- a/packages/date-picker/src/picker.vue
+++ b/packages/date-picker/src/picker.vue
@@ -741,9 +741,9 @@ export default {
       this.picker.selectionMode = this.selectionMode;
       this.picker.unlinkPanels = this.unlinkPanels;
       this.picker.arrowControl = this.arrowControl || this.timeArrowControl || false;
-      if (this.format) {
-        this.picker.format = this.format;
-      }
+      this.$watch('format', (format) => {
+        this.picker.format = format;
+      });
 
       const updateOptions = () => {
         const options = this.pickerOptions;
@@ -763,6 +763,11 @@ export default {
               option !== 'selectableRange') {
             this.picker[option] = options[option];
           }
+        }
+
+        // main format must prevail over undocumented pickerOptions.format
+        if (this.format) {
+          this.picker.format = this.format;
         }
       };
       updateOptions();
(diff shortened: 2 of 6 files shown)


Leopoldthecoder:
Hi, thanks for contributing!

Please add some test cases for the fix and new props.

remizovvv:
@Leopoldthecoder , hi!  I added two test cases, for `datetime` and `datetimerange`.

Leopoldthecoder:
@wacky6 What do you think of this PR? I'm not sure if we really need `date-format` and `time-format`.

wacky6:
I don't think such override is necessary, provided format extraction works correctly.

Having different format for panels and input-box seems odd.

remizovvv:
@Leopoldthecoder, @wacky6, Hi! Manual set `date-format` and `time-format` is an auxiliary functionality. Main functionality is a correct extracting date-format and time-format from format property. In the actual state this mechanism does not work (proof link to jsfiddle with this bug in my issue).

I can edit pullrequest, and remove `date-format` and `time-format` from `pickerOptions`. But main part of PR with bugfix will remain. 

What do you think about this?

wacky6:
Sound great.

Leopoldthecoder:
@remizovvv +1

remizovvv:
@wacky6, @Leopoldthecoder, OK :) I updated PR. 

In the process, I found another mini bug: if we set `format` property (not set `pickerOptions.format`), open picker panel, and next change`format` property, then format in picker panel will not change. Only set `pickerOptions.format` will update format in panel. You can try this:

https://jsfiddle.net/remizovvv/w2vrzru4/15/

I add fix for this in PR.

packages/date-picker/src/picker.vue
```diff
@@ -741,9 +741,12 @@ export default {
       this.picker.selectionMode = this.selectionMode;
       this.picker.unlinkPanels = this.unlinkPanels;
       this.picker.arrowControl = this.arrowControl || this.timeArrowControl || false;
-      if (this.format) {
-        this.picker.format = this.format;
-      }
+      this.$watch('format', (format) => {
+        // if main format is changed and pickerOptions.format undefined, then set to picker new main format
```
wacky6: pickerOptions.format is not documented (thus not publicly usable). I think you should always use main format.

packages/date-picker/src/util/index.js
```diff
@@ -215,3 +215,16 @@ export const nextYear = function(date, amount = 1) {
...
+export const extractDateFormatOnly = function(format) {
+  return format
+    .replace(/\W?m{1,2}|\W?ZZ/g, '')
+    .replace(/\W?h{1,2}|\W?s{1,3}|\W?a/gi, '')
+    .trim();
+};
+
+export const extractTimeFormatOnly = function(format) {
+  return format
+    .replace(/\W?D{1,2}|\W?Do|\W?d{1,4}|\W?M{1,4}|\W?y{2,4}/g, '')
```
wacky6: Looks like this function fails to strip `week` format.

image

packages/date-picker/src/panel/date.vue
```diff
@@ -153,7 +153,9 @@
     nextYear,
     prevMonth,
     nextMonth,
-    changeYearMonthAndClampDate
+    changeYearMonthAndClampDate,
+    extractDateFormatOnly,
```
wacky6: Can remove `Only` here. This word seems unnecessary.

packages/date-picker/src/picker.vue
remizovvv: Hmmm, pickerOptions.format not publicly usable, but user can set this option. I have seen this several examples in various issues. Do you think it will be correctly: if reseting main format will be overpowering a special pickerOptions.format? I think there can be a situation, when main format mustn't be equal picker panel format. For example, main format for short view, and picker format for advanced view. I think it's better to be safe. However, in two projects, where I use the element, there was no such need :)
What do you think?

wacky6: Resetting is not a good idea, it can cause timing issues. Might be better to your implementation at this stage if we come to a consensus that pickerOption.format is an issue.

The reason that setting pickerOption.format can change panel's format is historical: Picker simply proxies all properties in pickerOptions to underlying panel without any sanity check. (But it shouldn't)

My thought is: pickerOption.format is not documented (undefined behavior / not expected to be used by the user), setting it should fail the component in perceivable ways. So users know they are doing something wrong.

Otherwise it might create a false sense of "it works". This can cause unnecessary inconveniences if we choose to refactor internal implementations in the future. (See https://github.com/ElemeFE/element/issues/9878)

@Leopoldthecoder What do you think?

Leopoldthecoder: Agreed.

packages/date-picker/src/picker.vue
```diff
@@ -764,6 +764,10 @@ export default {
             this.picker[option] = options[option];
           }
         }
+
+        if (this.format) {
+          this.picker.format = this.format;
```
wacky6: I suggest adding a comment here noting this is to prevent user from setting undocumented `picker.format` (Like the one you added in the test).

I might be puzzling at first because there is a `$watch('format')` before.

wacky6: The rest of PR LGTM

remizovvv: @wacky6, good think, this some unobvious place. I added comment for this line.

Leopoldthecoder:
@remizovvv @wacky6 Thanks!

PR #10281, opened 2018-03-21
Scroll passive

kolesoffac:
Add options _{passive: true}_ for scroll event.


joshuaobrien:
What about browsers that do not support this feature?

Issue #10078, opened 2018-03-08
[Bug Report] default-time attribute only works with date range picker, not date picker


dak:
### Element UI version
2.2.1

### OS/Browsers version
All browsers

### Vue version
2.5.13

### Reproduction Link
[https://jsfiddle.net/mmx38qxw/535/](https://jsfiddle.net/mmx38qxw/535/)

### Steps to reproduce
1. Create a datetime picker with the default-time attribute
2. Choose a date
3. The time will be set by default to the current time, rather than the default-time

### What is Expected?
The default-time attribute should work with datetime picker the same way it works with datetimerange picker. When selecting a date, the time should be set to the value defined by the default-time attribute.

### What is actually happening?
The time is being set by default to the current time on the datetime picker.

Leopoldthecoder:
This not a bug. `default-time` only works in range mode.

@wacky6 Do you think we can get it to work in non-range mode?

wacky6:
Sounds ok to me. Should not be a huge problem.

I'd say PR welcome for this issue. I'm too busy recently to implement this myself.

Leopoldthecoder:
@wacky6 Got it.

gnijuohz:
@Leopoldthecoder I would like to work on this issue if no one is currently working on it :)

wacky6:
@gnijuohz Thanks for contribution. Just remember to write tests and update the documentation.

Leopoldthecoder:
Resolved in https://github.com/ElemeFE/element/pull/10321 .

PR #10321, opened 2018-03-22
DateTimePicker: Enable default-time attribute in date picker

RickMacTurk:
PR for issue: https://github.com/ElemeFE/element/issues/10078

### DIFF
--- a/examples/docs/en-US/datetime-picker.md
+++ b/examples/docs/en-US/datetime-picker.md
@@ -53,10 +53,11 @@
         },
         value1: '',
         value2: '',
-        value3: new Date(),
-        value4: '',
+        value3: '',
+        value4: new Date(),
         value5: '',
-        value6: ''
+        value6: '',
+        value7: ''
       };
     }
   };
@@ -117,6 +118,15 @@ DateTimePicker is derived from DatePicker and TimePicker. For a more detailed ex
       :picker-options="pickerOptions1">
     </el-date-picker>
   </div>
+  <div class="block">
+    <span class="demonstration">With default time</span>
+    <el-date-picker
+      v-model="value3"
+      type="datetime"
+      placeholder="Select date and time"
+      default-time="12:00:00">
+    </el-date-picker>
+  </div>
 </template>
 
 <script>
@@ -146,7 +156,8 @@ DateTimePicker is derived from DatePicker and TimePicker. For a more detailed ex
           }]
         },
         value1: '',
-        value2: ''
+        value2: '',
+        value3: ''
       };
     }
   };
@@ -163,7 +174,7 @@ DateTimePicker is derived from DatePicker and TimePicker. For a more detailed ex
   <div class="block">
     <span class="demonstration">Default</span>
     <el-date-picker
-      v-model="value3"
+      v-model="value4"
       type="datetimerange"
       range-separator="To"
       start-placeholder="Start date"
@@ -173,7 +184,7 @@ DateTimePicker is derived from DatePicker and TimePicker. For a more detailed ex
   <div class="block">
     <span class="demonstration">With shortcuts</span>
     <el-date-picker
-      v-model="value4"
+      v-model="value5"
       type="datetimerange"
       :picker-options="pickerOptions2"
       range-separator="To"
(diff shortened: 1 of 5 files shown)


examples/docs/en-US/datetime-picker.md
```diff
@@ -117,6 +118,15 @@ DateTimePicker is derived from DatePicker and TimePicker. For a more detailed ex
       :picker-options="pickerOptions1">
     </el-date-picker>
   </div>
+  <div class="block">
+    <span class="demonstration">With default time</span>
+    <el-date-picker
+      v-model="value3"
+      type="datetime"
+      placeholder="Select date and time"
+      :default-time="['12:00:00']">
```
wacky6: Why not a value but an array?

packages/date-picker/src/panel/date.vue
```diff
@@ -147,6 +147,7 @@
     isDate,
     modifyDate,
     modifyTime,
+    modifyWithGivenTime,
```
wacky6: perhaps `modifyWithTimeString` is a better name?

packages/date-picker/src/panel/date.vue
```diff
@@ -214,7 +215,10 @@
     methods: {
       proxyTimePickerDataProperties() {
         const format = timeFormat => {this.$refs.timepicker.format = timeFormat;};
-        const value = value => {this.$refs.timepicker.value = value;};
+        const value = value => {
+          const defaultTime = this.defaultTime || [];
```
wacky6: As mentioned before, for non-range, defaultTime should be a single value, not an array

RickMacTurk:
@wacky6 Made changes based on feedback.

Leopoldthecoder:
image

A test case has failed.

examples/docs/en-US/datetime-picker.md
```diff
@@ -117,6 +118,15 @@ DateTimePicker is derived from DatePicker and TimePicker. For a more detailed ex
       :picker-options="pickerOptions1">
     </el-date-picker>
   </div>
+  <div class="block">
+    <span class="demonstration">With default time</span>
+    <el-date-picker
+      v-model="value3"
+      type="datetime"
+      placeholder="Select date and time"
+      :default-time="'12:00:00'">
```
wacky6: How about `default-time="12:00:00"`? It is cleaner

packages/date-picker/src/util/index.js
```diff
@@ -142,6 +142,14 @@ export const modifyTime = function(date, h, m, s) {
   return new Date(date.getFullYear(), date.getMonth(), date.getDate(), h, m, s, date.getMilliseconds());
 };
 
+export const modifyWithDefaultTime = (date, time) => {
```
wacky6: How about `modifyWithTimeString`? DefaultTime does not imply argument type.

packages/date-picker/src/panel/date.vue
```diff
@@ -318,7 +321,7 @@
 
       handleDatePick(value) {
         if (this.selectionMode === 'day') {
-          this.date = modifyDate(this.date, value.getFullYear(), value.getMonth(), value.getDate());
+          this.date = modifyWithDefaultTime(value, this.defaultTime);
```
wacky6: This resets time to `defaultTime` whenever a date is picked. Current behavior sets time only if time is not set.

Comparison: 
* (current) https://jsfiddle.net/zbgmnqod/1/
* (this pr) https://jsfiddle.net/bmxe9u8r/

@Leopoldthecoder What's your thoughts on this behavior?

wacky6: Might need https://github.com/ElemeFE/element/issues/9878 (breaking changes) to resolve behavior differences of panels elegantly. 😂🤔

Leopoldthecoder: I think picking a date shouldn't change (reset) time.

examples/docs/en-US/datetime-picker.md
RickMacTurk: Made above change and resubmitted.

wacky6:
TODO:
* zh_CN, es docs
* update doc attribute's `type` & `accepted value` note

Leopoldthecoder:
I'll do the docs.

wuliupo:
it seems the default time still not work for date picker. (it is for good datetime picker)

wacky6:
it's for datetime not for date.

Issue #10433, opened 2018-03-29
[Bug Report] date-picker(type="datetime") compatible with the timestamp


xiongsongsong:
### Element UI version
2.3.0

### OS/Browsers version
all

### Vue version
2.5.17-beta.0

### Reproduction Link
[https://jsfiddle.net/xiongsongsong/mmx38qxw/1077/](https://jsfiddle.net/xiongsongsong/mmx38qxw/1077/)

### Steps to reproduce
```html
<el-date-picker type="datetime" v-model="val" / >
```

```javascript
data ( ) {
   return { val : Date.now() }
}
```

After access to the focus, the console will report an error.

### What is Expected?
Compatible with a timestamp.

### What is actually happening?
Not compatible with the timestamp.

wacky6:
Use `value-format='timestamp'`.

xiongsongsong:
Most of the time you don't know if it's a timestamp or a date type. Adding value-format is annoying.

like this: If (type="date") is ok, but it is (type="datetime") an error occurs.

it's ok.
```html
<el-date-picker
   v-model="timestamp"
   type="date"></el-date-picker>
```
it's wrong.
```html
<el-date-picker
   v-model="timestamp"
   type="datetime"></el-date-picker>
```

wacky6:
Binding timestamp is only permitted if value-format is `timestamp`. Otherwise it's undefined behavior, and may change in the future without warning.

My suggestion is not to rely on this kind of undocumented behavior. I also can't come up with scenarios where timestamp and date-string / date-object are mixed.

xiongsongsong:
40/5000
Hǎo ba.

Zhè míngmíng shì yīgè kěyǐ xiūbǔ de quēxiàn. Rúguǒ yāoqiú zhème yángé, wèishéme hái yǒu lèisì de pànduàn ne?
This obviously is a defect that can be repaired. If the requirements are so strict, why are there similar judgment?

https://github.com/ElemeFE/element/blob/dev/packages/date-picker/src/util/index.js#L28-L32
```
export const isDate = function(date) {
  if (date === null || date === undefined) return false;
  if (isNaN(new Date(date).getTime())) return false;
  return true;
};
```

xiongsongsong:
Too many similar scenes.

![image](https://user-images.githubusercontent.com/342509/38069016-c0e62d18-3346-11e8-8f3d-435c607955fc.png)

PR #10434, opened 2018-03-29
compatible with the timestamp

xiongsongsong:

#10433


wacky6:
Why is this change necessary. Time panel will never get non Date or non-null/undefined value

wacky6:
If your backend uses timestamp, you should use `value-format='timestamp'`.

PR #10498, opened 2018-03-31
Tooltip: invoke 'doDestory' by hand when disabled

blackmiaool:
`doDestroy` is not invoked when the tooltip is disabled because `<transition>` won't call `onAfterLeave`  when its content is not shown.
So `:popper-options="{removeOnDestroy:true}"` won't work when the tooltip is disabled.
https://jsfiddle.net/blackmiaool/3okg7754/4/

This pr solves this problem.

### DIFF
--- a/packages/tooltip/src/main.js
+++ b/packages/tooltip/src/main.js
@@ -191,6 +191,10 @@ export default {
         clearTimeout(this.timeoutPending);
       }
       this.showPopper = false;
+
+      if (this.disabled) {
+        this.doDestroy();
+      }
     },
 
     setExpectedState(expectedState) {


packages/tooltip/src/main.js
```diff
@@ -191,6 +191,10 @@ export default {
         clearTimeout(this.timeoutPending);
       }
       this.showPopper = false;
+
+      if (this.disabled) {
```
Leopoldthecoder:
```js
if (this.disabled && this.popperOptions.removeOnDestroy) {
```

blackmiaool: is it necessary?

Leopoldthecoder: Yes, we should only remove DOM when `this.popperOptions.removeOnDestroy` is true. Otherwise, the user may have some DOM manipulation even when the Tooltip is hidden, and `doDestroy` here may lead to some unexpected errors.

blackmiaool: > we should only remove DOM when this.popperOptions.removeOnDestroy

It is implemented inside "popper.js"  https://github.com/ElemeFE/element/blob/dev/src/utils/popper.js#L213

Leopoldthecoder: Ah, you're right.

badpunman:
this change introduced a memory leak
https://github.com/ElemeFE/element/issues/19370

PR #10623, opened 2018-04-09
Pagination: handle NaN on props

brandon-2345:

Pagination component breaks when `NaN` is passed on the prop `currentPage` or `pageSize`. Here is a demo reproducing the issue.

https://jsfiddle.net/xroupdkf/2/
screen shot 2018-04-09 at 5 48 11 pm

I added a little protection code to keep it safe from `NaN`.
- With `NaN`, `pageSize` returns default value (which is `1`).
- `currentPage` already had a nice handling code inside `internalCurrentPage` watcher but it is not invoked at initial phase, so I made it to be called immediately.

It passes all of the test cases and I added more of it.

Thank you for maintaining this amazing repository.

### DIFF
--- a/packages/pagination/src/pagination.js
+++ b/packages/pagination/src/pagination.js
@@ -379,23 +379,26 @@ export default {
       }
     },
 
-    internalCurrentPage(newVal, oldVal) {
-      newVal = parseInt(newVal, 10);
+    internalCurrentPage: {
+      immediate: true,
+      handler(newVal, oldVal) {
+        newVal = parseInt(newVal, 10);
 
-      /* istanbul ignore if */
-      if (isNaN(newVal)) {
-        newVal = oldVal || 1;
-      } else {
-        newVal = this.getValidCurrentPage(newVal);
-      }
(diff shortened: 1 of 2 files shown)


packages/pagination/src/pagination.js
```diff
@@ -375,27 +375,30 @@ export default {
     pageSize: {
       immediate: true,
       handler(val) {
-        this.internalPageSize = val;
+        this.internalPageSize = isNaN(val) ? 10 : val;
```
Leopoldthecoder: Hi, I think it should be `1`, right?

brandon-2345: Hey! I though it was `10` because `pageSize`'s default value is `10` as written on prop type definitions.

https://github.com/ElemeFE/element/blob/6f4f57172b945f7f94563eb11358bb535e5639c6/packages/pagination/src/pagination.js#L14

Am I missing something? Sorry, I'm kind of new to this part.

Leopoldthecoder: Yeah you're right.

PR #10842, opened 2018-04-23
Select: fix multiselect tag text overflow display bug

yugasun:
Bug Reproduction Link
https://jsfiddle.net/yugasun/11wehv40/


dborstelmann:
Would be great if this got merged!

PR #10939, opened 2018-04-28
[Button] Add margin for button-group + button

dmitriyshmatov:
Hi, made some minor edits.


Leopoldthecoder:
This may not be desired by every user. You can fine tune any style using CSS in you own project.

PR #11040, opened 2018-05-07
Input-number: add readonly attribute

asmyshlyaev177:


asmyshlyaev177:
@Leopoldthecoder  If need to disable control buttons and input area have attribute 'disabled'
Whole point is ability to deny edit input field from keyboard.
Because when user can type non digit symbols or remove value when "min" attribute used it is look strange.

Leopoldthecoder:
Actually `readonly` means you cannot edit the value at all. Look how native input works: https://jsfiddle.net/2otvjrz8/ . With `readonly`, the control buttons are hidden.

asmyshlyaev177:
Hm, i see, maybe just rename attribute, input-readonly, only-controls or input disabled?

Leopoldthecoder:
Then I suggest you manipulate DOM directly: https://jsfiddle.net/jwmyb8ek/ , as this doesn't seem to be a common use case.

PR #11210, opened 2018-05-17
Upload: fix a TypeError when upload a processed file from the `beforeUpload` hook

fengyuanchen:

### DIFF
--- a/packages/upload/src/upload.vue
+++ b/packages/upload/src/upload.vue
@@ -93,10 +93,9 @@ export default {
         before.then(processedFile => {
           const fileType = Object.prototype.toString.call(processedFile);
 
-          processedFile.name = rawFile.name;
-          processedFile.uid = rawFile.uid;
-
           if (fileType === '[object File]' || fileType === '[object Blob]') {
+            processedFile.name = rawFile.name;
+            processedFile.uid = rawFile.uid;
             this.post(processedFile);
           } else {
             this.post(rawFile);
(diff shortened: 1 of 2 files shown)


Leopoldthecoder:
Please write a demo of the bug you're trying to fix.

fengyuanchen:
Here is a demo: https://jsfiddle.net/cfy/53jf2oms/1/

Open the console and choose an image to upload, and then you will see a TypeError in the console.

fengyuanchen:
In line 56 of **ajax.js**, I also add the file name to the form data, just because if miss it, the server side will get a file with the name of `'blob'` (without extension) by default after upload the compressed image file, but we may want a valid name as`'example.jpg'`.

fengyuanchen:
In line 84 of **upload.vue**, the second argument is not used, so I remove it.

packages/upload/src/upload.vue
```diff
@@ -92,6 +92,10 @@ export default {
       if (before && before.then) {
         before.then(processedFile => {
           const fileType = Object.prototype.toString.call(processedFile);
+
+          processedFile.name = rawFile.name;
```
Leopoldthecoder: Is this line necessary? Because here `processedFile` already has a `name` prop.

fengyuanchen: A [blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob) object only has two properties: `size` and `type`. In the given demo, the `name` property is assigned by the image compressor.

Leopoldthecoder: OK

qusiba: This change will cause an error if here the processedFile is still a File object.

PR #11291, opened 2018-05-22
Added `noChange` parameter to `toggleRowExpansion` due to infinite recursion problems

galvez:
In order to solve the issue without the proposed parameter, one needs to add a global flag and check manually on the handler for `expand-change`. I feel this should be added to the API as it's simple to achieve.


galvez:
@Leopoldthecoder check this out when you can :)

galvez:
@Leopoldthecoder the build test failure seems unrelated to the code I touched, something about a `chai` return value elsewhere. I've seen `chai` related build failures happen before. Can you check? I think this PR is ok (the first test passed).

Leopoldthecoder:
Thanks for contributing.

Actually `expand-change` should never trigger when `toggleRowExpansion` is called, because all these 'change' events only trigger on user input.

Considering this is a breaking change, we'll take care of it in a major version.

Issue #11450, opened 2018-06-01
[Bug Report] Datetime Picker with `disabledDate` doesn't work as expect


plucury:
### Element UI version
2.4.0

### OS/Browsers version
mac 10.13.3 / Chorme  66.0.3359.181（正式版本） （64 位）

### Vue version
2.5.16

### Reproduction Link
https://codesandbox.io/s/v8283jzv00

### Steps to reproduce
1. select a  valid date
2. click on a invalid date(the month of the date should before the mouth of the valid date)
3. select a time

### What is Expected?
The date should be the one we selected at step 1.

### What is actually happening?
The date is the one we clicked at step 2.

wacky6:
It's unrelated to disabledDate. It's actually caused by using incorrect internal variable when time is picked. https://github.com/ElemeFE/element/pull/11474

Leopoldthecoder:
Fixed in #11474

PR #11474, opened 2018-06-04
date-time-picker: fix incorrect date on time pick

wacky6:
Regression introduced during 2.0 refactor. Quite surprised that it took 7 months for someone to spot this bug. 😂

`modifyWithDefaultTime` changed to `modifyWithTimeString` to reflect what it does.

### DIFF
--- a/packages/date-picker/src/panel/date-range.vue
+++ b/packages/date-picker/src/panel/date-range.vue
@@ -190,7 +190,7 @@
     isDate,
     modifyDate,
     modifyTime,
-    modifyWithDefaultTime,
+    modifyWithTimeString,
     prevYear,
     nextYear,
     prevMonth,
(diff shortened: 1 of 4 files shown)


Leopoldthecoder:
Actually I think what #11450 meant was

untitled1

wacky6:
Updated, I missed the `confirm()` bit. It should work correctly now.
https://jsfiddle.net/4frk3fmd/

---

Recommend CI/bot reply to include url of compiled element-ui bundle that can be used to write jsfiddle happily. It might be troublesome if amendments are made (thus changing bundle's hash -> http 404). Right now, I'm happy with my [js-bin](https://github.com/wacky6/js-bin) hack. :D

Leopoldthecoder:
It seems the compiled `element-ui.[hash].js` for netlify / official website cannot be used in jsfiddle. @jikkai could you verify this?

zaoyu63:
this.value 如果不是日期类型，是个字符串， modifyTime方法就会报错!!!
This value if not date type, is a string, modifyTime method will be an error!!!

jayzch:
重构有问题啊！！问题同上，整个系统挂了。

wacky6:
https://github.com/ElemeFE/element/issues/11621#issuecomment-398272178

PR #11515, opened 2018-06-06
Input: Do not call calcTextareaHeight if unnecessary for performance

suusan2go:

Calling `scrollHeight` forces layout / reflow and sometimes causes serious performance issue (especially at Edge, in my environment).

image

`Input` component having type `textarea` calls `calcTextareaHeight ` by every input event. 
https://github.com/ElemeFE/element/blob/4b3edc13abbac373d58503f9bc5223a7853d6df0/packages/input/src/input.vue#L206-L220
https://github.com/ElemeFE/element/blob/4b3edc13abbac373d58503f9bc5223a7853d6df0/packages/input/src/input.vue#L224-L228

And `calcTextareaHeight` calls  `scrollHeight` everytime at this point.
https://github.com/ElemeFE/element/blob/4b3edc13abbac373d58503f9bc5223a7853d6df0/packages/input/src/calcTextareaHeight.js#L73

I think it's unecessary to recalculate `textarea` minHeight for every input event, if autosize option is false. Because minHeight will not change in most usecases.

This change improves Input component(having type `textarea` and not having `autosize` option)  performance.

FYI: https://gist.github.com/paulirish/5d52fb081b3570c81e3a

### DIFF
--- a/packages/input/src/input.vue
+++ b/packages/input/src/input.vue
@@ -245,7 +245,7 @@
         if (type !== 'textarea') return;
         if (!autosize) {
           this.textareaCalcStyle = {
-            minHeight: calcTextareaHeight(this.$refs.textarea).minHeight
+            minHeight: this.textareaCalcStyle.minHeight || calcTextareaHeight(this.$refs.textarea).minHeight
           };
           return;
         }


arvidkahl:
Please revisit this pull request. This performance issue has been hitting our application hard. The fix would be a very welcome addition to the library.

suusan2go:
@arvidkahl For a workaround for this issue, I made a custom component like this.

```js
import Vue from 'vue';
import calcTextareaHeight from './calcTextareaHeight.js'; // just copy from element-ui

const ElInput = Vue.options.components['ElInput'];

export default ElInput.extend({
  methods: {
    // override
    resizeTextarea() {
      if (this.$isServer) return;
      const { autosize, type } = this;
      if (type !== 'textarea') return;
      if (!autosize) {
        this.textareaCalcStyle = {
          minHeight:
            this.textareaCalcStyle.minHeight || calcTextareaHeight(this.$refs.textarea).minHeight,
        };
        return;
      }
      const minRows = autosize.minRows;
      const maxRows = autosize.maxRows;

      this.textareaCalcStyle = calcTextareaHeight(this.$refs.textarea, minRows, maxRows);
    },
  },
});

```

wacky6:
@Leopoldthecoder Could you take a look at why [input.vue#L210-215](https://github.com/ElemeFE/element/blame/4b3edc13abbac373d58503f9bc5223a7853d6df0/packages/input/src/input.vue#L210-L215) was added?

nicooprat:
Looks like this function takes ~16ms for every character in IE11: https://github.com/ElemeFE/element/blob/f6df39e8c1ff390da0f0df8ea30b07baf5d457f0/packages/input/src/calcTextareaHeight.js#L31

Fixed it using [autosize.js](https://github.com/jackmoore/autosize) and the hack above:

```js
// Textarea.vue
<script>
import { Input } from 'element-ui';
import autosize from 'autosize';

export default {
  extends: Input,
  mounted() {
    if (this.autosize) {
      autosize(this.$refs.textarea);
    }
  },
  destroyed() {
    if (this.autosize) {
      autosize.destroy(this.$refs.textarea);
    }
  },
  methods: {
    resizeTextarea() {},
  },
};
</script>
```

Then you can import it locally `import Textarea from '@/components/Textarea';` and/or globally `Vue.component('ElTextarea', Textarea);`

Finally use it like `el-input` but with a simpler boolean `autosize` prop and the standard `rows` attribute:

```html
<el-textarea type="textarea" rows="4" autosize />
```

So the only downside is that we can't have min/max rows parameters in this case.

I hope this get resolved as soon as possible because it looks like **every** input (not only textareas with or without autosize) gets this lag on Internet Explorer 11 which makes any app pretty unusable.

mikemand:
For some reason, Element's inputs cause Vue to re-compute directives:

Screen Shot 2020-01-20 at 5 49 23 PM

Any time text is input, it would re-run directive(s) for each character typed. I wasn't able to directly see where Element was causing this, but I don't have input lag on another Vue site that doesn't use Element. I haven't tried my directive to see if that was the cause, but I will (and I'll edit this if it turns out it was my directive and not Element).

ke1vin1993:
Why this PR is still open? Textarea component will calculate height when it mounted, this could cause reflow when rendering lot of Textarea and lead to serious performance issues. I almost give up Element-ui.

nicooprat:
Just found an even easier override:

```vue
<script>
import { Input } from 'element-ui';

export default {
  extends: Input,
  methods: {
    // Override to avoid performance issues
    // https://github.com/ElemeFE/element/pull/11515
    resizeTextarea() {
      this.$refs.textarea.parentElement.style.height = this.$refs.textarea.style.height;
      this.$refs.textarea.style.height = '';
      this.$refs.textarea.style.height = `${this.$refs.textarea.scrollHeight}px`;
      this.$refs.textarea.parentElement.style.height = '';
    },
  },
};
</script>

<style scoped>
textarea {
  resize: none;
  overflow: hidden;
}
</style>
```

Seems to work even on IE 11 👍

PR #11559, opened 2018-06-09
FormItem: Pass full form model to form item validation

st-sloth:
In some cases a form field needs complex validation logic that depends on other fields in the form (for example, a form has a country select and a phone number / bank card input that is validated by selected country). So the custom `validator` function needs access to the full form model.
The `descriptor` object contains only rules for the current field anyway, so extra fields on the validated model won't affect the validation.

### DIFF
--- a/packages/form/src/form-item.vue
+++ b/packages/form/src/form-item.vue
@@ -190,7 +190,7 @@
         descriptor[this.prop] = rules;
 
         const validator = new AsyncValidator(descriptor);
-        const model = {};
+        const model = objectAssign({}, this.form.model);
 
         model[this.prop] = this.fieldValue;
 


ziyoung:
Thanks for your pr. We can get values of other fields in custom function `validator`. http://element.eleme.io/#/en-US/component/form#custom-validation-rules  I think there is no need to do this.

st-sloth:
> We can get values of other fields in custom function `validator`.

This is exactly what this PR helps with. [The provided example](http://element.eleme.io/#/en-US/component/form#custom-validation-rules) only has access to the other fields of the form because it is specifically declared in such a place, so there is direct access to the object through a closure. If these validation functions where not in the `data()` or the `methods` but, for example, at the same level as the component's declaration, that wouldn't work.

Also, in the case of a "generalized form builder", the configuration (including `validator`) gets passed as props from outside, and there is no reference to the form or the model there. So one needs the model reference directly as the `validator` argument, [which is the 4th one](https://github.com/yiminghe/async-validator#rules).

st-sloth:
Hi, @ziyoung, please, give it another look :pray: . It is a must for my case and I have to update my fork, which is a one-line change, to keep up with this great library.

wacky6:
You can use high-order function to encapsulate whatever information you need to the validator.

st-sloth:
Indeed, a high-order function helps (though a bit ugly), thank you!
However, wouldn't it be logical to have the full form model data passed to [the provided `async-validator` API](https://github.com/yiminghe/async-validator#rules)? Beside from it being a breaking change for those who based their code on the fact that currently `source` is an object with the single field being validated. I understand that this is a valid concern. I will cancel the PR, just curious about your opinion on that.

st-sloth:
Thank you for the detailed explanation! I see your point.
And the way I understand

> The use of high-order function here is to **convey the concept that A depends on B**.

you mean the HOF has access to the model, to return proper validation based on value of the dependency, and is used in the scope of the validation rule.

But I don't see how that is applicable if the model is stored levels deeper in the component tree than the validation rules, so one doesn't have access from the validators to the model (or different models, when the same field validators used in a creation form and a modification form, that have different models), even transitively through a HOF.

The way I used your advice is actually hidden from the validator scope so there is no HOF to convey that A depends on B. I have a form generator component (that uses `el-form`) that gets form configuration (including validation rules) and the model from the props. And inside this component I substitute passed validators with wrappers that get the first 3 arguments from the internals of `el-form` / `el-form-item`, and call the passed ones with those and the model prop as the 4th, basically giving outside validators full model as the 4th argument instead of an object with a single property.

Yes, that is ugly and probably not the way you meant for a high-order function to be used. But a validation rule for A is based on a value of B (like validating a phone number or a town based on a country) and B's value isn't available in the scope of the validator. So the `source` argument in the validator is the only hope and that monkey patching deeper in the component chain solves the problem.
While a bit different, you still pushed be to a solution, thank you!

longshihui:
I have same problem....

PR #11762, opened 2018-06-26
message-box: close/cancel

gdi2290:
We needed 3 states for our confirm. In Tipe if you make an edit and leave the page we ask if they want to Cancel (discard changes) or Success (save and continue). We needed a Close state for when they click on [x] or click out of the modal to close the modal without doing anything
* Success
* Cancel
* Close

### DIFF
--- a/packages/message-box/src/main.vue
+++ b/packages/message-box/src/main.vue
@@ -22,8 +22,8 @@
             class="el-message-box__headerbtn"
             aria-label="Close"
             v-if="showClose"
-            @click="handleAction('cancel')"
-            @keydown.enter="handleAction('cancel')">
+            @click="handleAction('close')"
+            @keydown.enter="handleAction('close')">
             <i class="el-message-box__close el-icon-close"></i>
           </button>
         </div>
@@ -173,7 +173,7 @@
 
       handleWrapperClick() {
         if (this.closeOnClickModal) {
-          this.handleAction('cancel');
+          this.handleAction('close');
         }
       },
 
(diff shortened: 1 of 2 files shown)


ziyoung:
@gdi2290 Is it necessary to add `close` status? 
We usually use `message-box` as follows:
```js
this.$messageBox('title', 'content').then(callback).catch(callback)
```
The `confirm` status and `cancel` status correspond to the status of Promise. It seems there's no need to add one more status.

gdi2290:
@ziyoung yeah for our needs we named the cancel button something else so we reactive to the user depending on 3 states (confirm, cancel, close). If we had a way to add another success button then we can also use that so the catch can remain the cancel. It would be better to provide different success calls 

Here's our use-case
screen shot 2018-06-26 at 8 35 18 pm
The [x] is the same as discord changes but it would be better to cancel the model/route change.

Here's the code vue.js/nuxt.js that deals with the 3 states in the router leave lifecycle hook.
screen shot 2018-06-26 at 8 37 26 pm
So we either need to add more error options or add more success options

Leopoldthecoder:
How about adding a prop to decide if clicking on [x] or clicking out of the modal are considered to be `close` instead of `cancel`? And this prop should default to `false` for backward compatibility.

gdi2290:
yeah I can work with that

ziyoung:
@gdi2290 #11831 `distinguishCancelAndClose` attribute has been added to MessageBox.

gdi2290:
thanks!

HungryBird:
why it dosent work when I copy demo's code on http://element-cn.eleme.io/#/zh-CN/component/message-box.
I missing what type?

PR #12047, opened 2018-07-18
Table: add toggleAllSelection method

### LINKED ISSUE #11978: [Feature Request] Table component should append on-change hook and prop "default-select-all"
### Existing Component
是

### Component Name
Table

### Description
Yes, It was very hard and trouble to handler when My Table component was very complex， Because of I want to render multiple tables with similar type dataes ，I just want default-select-all  but the component didnt support !!!  I didnt want to use  toggleRowSelection again！！！

jikkai:
Resolves #11978


weituotian:
this method has a problem,  it can select all unselected row,  but previou selected row can not change to be unselect

PR #12061, opened 2018-07-19
fixed clearable input with append item

whthT:

video
https://www.youtube.com/watch?v=AycT46fGZF4

### DIFF
--- a/packages/input/src/input.vue
+++ b/packages/input/src/input.vue
@@ -132,7 +132,7 @@
         focused: false,
         isOnComposition: false,
         valueBeforeComposition: null,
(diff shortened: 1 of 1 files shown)


ziyoung:
https://jsfiddle.net/zhiyang/xyu2jbv7/
The problem only occurs when `ElInput` contains `ElSelect`. We use a function called `calcIconOffset` to calculate offset. Try to edit `calcIconOffset`.
https://github.com/ElemeFE/element/blob/43f83128cdc16fa6c4d2c698d3b15b674733688e/packages/input/src/input.vue#L293-L307

ziyoung:
Fix in #12108

PR #12076, opened 2018-07-20
Tree: fix a bug dropNode doesn't change background when the dropNode is the grandfather's father of the dragging node

beryl20120308:
when I drag a node to its grandfater's father, the background of the grandfather's father is expected to be blue, but the background don't change.  
I add the dropType to the emit event "node-drag-over" and change the background in my project by the "node-drag-over" and "node-drag-leave" event.

### DIFF
--- a/packages/tree/src/tree.vue
+++ b/packages/tree/src/tree.vue
@@ -438,7 +438,7 @@
         dragState.showDropIndicator = dropType === 'before' || dropType === 'after';
         dragState.allowDrop = dragState.showDropIndicator || userAllowDropInner;
         dragState.dropType = dropType;
-        this.$emit('node-drag-over', draggingNode.node, dropNode.node,dropType, event);
+        this.$emit('node-drag-over', draggingNode.node, dropNode.node, dropType, event);
       });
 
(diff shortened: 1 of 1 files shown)


ziyoung:
@beryl20120308 please provide a demo and explain why add `dropType`.

beryl20120308:
I take the draggable Tree on the website  [I](http://element-cn.eleme.io/#/en-US/component/tree)  for example.when I drag the Level three 3-2-1 to the Level two 2-1, the background of the level two 2-1 will go blue, but when I drag it to the Level one 2 who is the level two 2-1’s parent ,the background of the level one 2 won’t change.but actually the Level there 3-2-1 can be dragged into the Level one 2. And If the level one 2 has parent, If I drag one node to it, the thing also happens,and so on. I found when I drag the node the event “drag-over” is triggered, so if I know the dropType, I can control the background myself. maybe there is better solution for this, but I solve it by adding the variable of dropType. @ziyoung

beryl20120308:
@ziyoung I have changed the source code in my project and everything goes well. But I have no idea when I can publish my project because I m not sure when my code can be merged and when I can use the code I have changed.

PR #12164, opened 2018-07-30
Tooltip: improved the readability.

k-okina:
The code at the point where popperVM is generated should be separated by an easy-to-understand function etc.
Because easy-to-understand function names replace comments and maintainability is improved.


ziyoung:
The function is not reused. Is it necessary to do this?

jikkai:
Thanks for your PR!

In my opinion, it is unnecessary to create an intermediate variable for creating popperVM. 

Plus:

* `generatePopperVM` function has no need to be reused  in other contexts.
* `beforeCreate` hook is simple enough to understand.

So I think the current code doesn't break the readability.

k-okina:
I appreciate your review!
That's right.
I close this pull request!

PR #12239, opened 2018-08-06
Menu: make font-size constant into variable

### LINKED ISSUE #12238: [Feature Request] font-size in scss variables?
### Existing Component
yes

### Component Name
Menu
path: `element-ui/packages/theme-chalk/src/menu.scss` `line: 9`

### Description
Please use the scss variable to set the font-size so that we can adjust the font size as a whole.

jikkai:
Resolves #12238


shellus:
so fast

PR #12241, opened 2018-08-06
Autocomplete: add `popper-append-to-body` attribute

### LINKED ISSUE #12211: [Feature Request] Implement popperAppendToBody in autocomplete component
### Existing Component
Yes

### Component Name
autocomplete

### Description
It's required if the input is in a component which close itself on `outsideClick`, I think that for convenience, all the components should have `popperAppendToBody` parameter that passes `append-to-body` to the inside popover.

ziyoung:
Closes #12211


SherryJiang1130:
my element-ui have updated to 2.4.6  but popper-append-to-body='false' still not work

Issue #6754, opened 2017-08-29
[Bug Report] datepicker week type doesn't consider firstWeekOfDay


AshwinTayson:
### Element UI version
1.4.3

### OS/Browsers version
Ubuntu 16.04 / Chrome Latest

### Vue version
2.4.2

### Reproduction Link
[https://jsfiddle.net/pj71jkyw/517/](https://jsfiddle.net/pj71jkyw/517/)

### Steps to reproduce
1.Click on the date picker.
2.Pick any week

### What is Expected?
It should take Sunday as the first day since it has prop firstDayOfWeek=7

### What is actually happening?
By default it takes monday for any value of firstDayOfWeek

Leopoldthecoder:
`firstDayOfWeek` should be nested in `picker-options`: https://jsfiddle.net/pj71jkyw/520/

AshwinTayson:
It doesn't selected sunday or any other other day even after the the change. It is monday by default.

pushchris:
@Leopoldthecoder I am experiencing the issue as well, your given JSFiddle actually demonstrates the issue. If the ```firstDayOfWeek``` parameter is set to 7 it should be selecting the date of Sunday instead of Monday. The issue seems to be caused by this line: 

https://github.com/ElemeFE/element/blob/762f337b2c8eac32eb1cc85c4af1e7718973cdeb/packages/date-picker/src/basic/date-table.vue#L364-L366

Where it is automatically selecting the cell at index 1. Shouldn't this change depending on what the ```firstDayOfWeek``` parameter is? A better option might actually be to select the first cell (index 0) no matter what so that when you do set the ```firstDayOfWeek``` so you know it will be returning that day. Currently in your example when you have the first day of the week set to Wednesday, it is returning the date for Thursday. I believe what @AshwinTayson is wanting is for it to return the day of the week she passes in.

Leopoldthecoder:
Yes, this is a bug. Will take a look later.

simshaun:
So that's the reason why it always chooses the day after the firstDayOfWeek here?
https://jsfiddle.net/simshaun/mmx38qxw/1927/

firesh:
Hi all, this issue actually still not fixed. I raise a new issue with repro link.
#15683

Toukassay:
Hello, still no fix, right ?

jasonchruy:
still no fix?

PR #12320, opened 2018-08-10
DatePicker: in week mode set value to first day of week when user click on cell

elfman:
fix issue #6754

### DIFF
--- a/packages/date-picker/src/basic/date-table.vue
+++ b/packages/date-picker/src/basic/date-table.vue
@@ -400,7 +400,7 @@
         const selectionMode = this.selectionMode;
 
         if (selectionMode === 'week') {
-          target = target.parentNode.cells[1];
+          target = target.parentNode.cells[0];
         }
 
         let year = Number(this.year);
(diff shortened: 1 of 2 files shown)


wacky6:
This is a breaking change, and should be merged into `next`.

wacky6:
Also, in ISO-8601, first day of week is Monday.

Given the current default date-table starts a week from Sunday. After this fix, the component will emit Sunday's date. I doubt if this behavior is what the user really wants.

elfman:
@wacky6 The thing you doubt about should be another case.
Setting the default value of `firstDayOfWeek` to `1` can make `DatePicker` behave as ISO-8601

wacky6:
Setting `firstDayOfWeek = 1` will change the date-table layout, But the user may want to have Sunday to be displayed as the first day, and have the value to start from Monday.

elfman:
@liwenda122 能否说一下哪个地方还有问题？

island205:
Any update?

wacky6:
IMO, week picker should eventually emit a date range. i.e. ['2018-02-03', '2018-02-09'].

Current behavior emits a single date to represent the entire week.

---

This pr breaks current default behavior, which is to have Sunday as the first day on date-table, and emit Monday as component's value.

Issue #11392, opened 2018-05-29
[Bug Report] Writing space become small when collapse-tags prop is added


akki-jat:
### Element UI version
2.4.0

### OS/Browsers version
Windows 7 / Chrome  66.0.3359.181

### Vue version
2.5.11

### Reproduction Link
[https://codepen.io/akki-jat/pen/mLqzPV](https://codepen.io/akki-jat/pen/mLqzPV)

### Steps to reproduce
Write few words in select component to check writing space

### What is Expected?
Same writing space even if collapse-tags prop is added

### What is actually happening?
Writing spaced is different if collapse-tags prop is added

PR #12329, opened 2018-08-10
Select: Fix (#11392 & #12327)

### LINKED ISSUE #12327: [Bug Report] input takes new line even if space is available when filterable & multiple props are true
### Element UI version
2.4.6

### OS/Browsers version
Windows 7 / Chrome 68.0.3440.106 (Official Build) (64-bit)

### Vue version
2.5.2

### Reproduction Link
[https://jsfiddle.net/akki_jat/94ba3s2p/1/](https://jsfiddle.net/akki_jat/94ba3s2p/1/)

### Steps to reproduce
Just select three options and you will see input takes the whole new line.

Note: Select 3 options

### What is Expected?
Use the available space

Taking whole new line

### Screenshot
element_issue

akki-jat:
Fix for Writing space become small when collapse-tags prop is added & input takes new line even if space is available when filterable & multiple props are true

- Closes #12329 
- Closes #12327

### DIFF
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -28,7 +28,7 @@
           <span class="el-select__tags-text">+ {{ selected.length - 1 }}</span>
         </el-tag>
       </span>
-      <transition-group @after-leave="resetInputHeight" v-if="!collapseTags" style="display: contents;">
+      <transition-group @after-leave="resetInputHeight" v-if="!collapseTags">
         <el-tag
           v-for="item in selected"
           :key="getValueKey(item)"
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -28,7 +28,7 @@
           <span class="el-select__tags-text">+ {{ selected.length - 1 }}</span>
         </el-tag>
       </span>
-      <transition-group @after-leave="resetInputHeight" v-if="!collapseTags">
+      <transition-group @after-leave="resetInputHeight" v-if="!collapseTags" style="display: contents;">
         <el-tag
           v-for="item in selected"
           :key="getValueKey(item)"
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -8,7 +8,7 @@
       class="el-select__tags"
       v-if="multiple"
       ref="tags"
-      :style="{ 'max-width': inputWidth - 32 + 'px' }">
+      :style="{ 'max-width': inputWidth - 32 + 'px', width: '100%' }">
       <span v-if="collapseTags && selected.length">
         <el-tag
           :closable="!selectDisabled"
@@ -28,7 +28,7 @@
           <span class="el-select__tags-text">+ {{ selected.length - 1 }}</span>
         </el-tag>
       </span>
-      <transition-group @after-leave="resetInputHeight" v-if="!collapseTags">
+      <transition-group @after-leave="resetInputHeight" v-if="!collapseTags" style="display: contents;">
         <el-tag
           v-for="item in selected"
           :key="getValueKey(item)"
@@ -64,7 +64,7 @@
         v-model="query"
         @input="debouncedQueryChange"
         v-if="filterable"
-        :style="{ width: inputLength + 'px', 'max-width': inputWidth - 42 + 'px' }"
+        :style="{ 'flex-grow': '1', width: '30%' }"
         ref="input">
     </div>
     <el-input
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -64,7 +64,7 @@
         v-model="query"
         @input="debouncedQueryChange"
         v-if="filterable"
-        :style="{ 'flex-grow': '1', width: '30%' }"
+        :style="{ 'flex-grow': '1', width: inputLength / 208 + '%', 'max-width': inputWidth - 42 + 'px' }"
         ref="input">
     </div>
     <el-input
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -64,7 +64,7 @@
         v-model="query"
         @input="debouncedQueryChange"
         v-if="filterable"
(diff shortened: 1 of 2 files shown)


packages/select/src/select.vue
```diff
@@ -28,7 +28,7 @@
           <span class="el-select__tags-text">+ {{ selected.length - 1 }}</span>
         </el-tag>
       </span>
-      <transition-group @after-leave="resetInputHeight" v-if="!collapseTags">
+      <transition-group @after-leave="resetInputHeight" v-if="!collapseTags" style="display: contents;">
```
jikkai: IE & Edge doesn't support `display: contents` 🤔

akki-jat:
@jikkai changes are done as you requested. Can you review now?

packages/theme-chalk/src/select.scss
```diff
@@ -109,7 +109,6 @@
     z-index: $--index-normal;
     top: 50%;
     transform: translateY(-50%);
-    display: flex;
     align-items: center;
```
ziyoung: The following two lines can be deleted.

packages/select/src/select.vue
```diff
@@ -64,7 +64,7 @@
         v-model="query"
         @input="debouncedQueryChange"
         v-if="filterable"
-        :style="{ width: inputLength + 'px', 'max-width': inputWidth - 42 + 'px' }"
```
ziyoung: When `collapse-tags` is set to `true`, there are some problems.
kapture 2018-08-16 at 11 25 36

Maybe you should not modify this line.

akki-jat: @ziyoung now check.

akki-jat:
@jikkai I understand there is no support for **display: contents** in IE & Edge, but if I put display: contents at-least it **behaves good** in **chrome & firefox** and for **IE/Edge fallback to old behaviour**.

So I think its good to have **correct behaviour in two popular browsers** rather than having **faulty behaviour in all browsers**.

What do you think? Tell me.

akki-jat:
@ziyoung @jikkai any response?

akki-jat:
@ziyoung @jikkai any response?

ziyoung:
@akki-jat Your pr does fix these issues. But I don't know whether this will create new problems.

akki-jat:
@ziyoung do some testing, if you want. And I am telling you if anything comes regarding this I will be happy to help. But do not hesitate to fix these issues.

wacky6:
LGTM. Nice to have progressive enhancements.

I don't mind breaking IE, since Vue 3.0 is likely to break it through use of Proxy anyway. It won't hurt to add modern browser only features, so that we can push Element users to convince their clients to use modern browsers.

akki-jat:
@wacky6 This commit is not breaking IE, it will just just behave like it is behaving right now

packages/select/src/select.vue
ziyoung: @akki-jat have you fixed it?
kapture 2018-10-09 at 15 36 08

packages/select/src/select.vue
```diff
@@ -64,7 +64,7 @@
...
+        :style="{ 'flex-grow': '1', width: inputLength / 208 + '%', 'max-width': inputWidth - 42 + 'px' }"
```
ziyoung: If I don't modify this line, I find it also works fine. By the way, what does 208 mean?

packages/select/src/select.vue
akki-jat: @ziyoung I have fixed it. Why? you still getting this behaviour?

packages/select/src/select.vue
akki-jat: 208 is the width of el-select__tags (I know it should be dynamic and I am pushing new commit to make it dynamic)
& What do you mean by If I don't modify this line. (Not working with updated line?)

akki-jat:
@ziyoung updated 208 (static width). Check now. And anything comes let me know.

packages/select/src/select.vue
akki-jat: @ziyoung No, this was just **Error Reproduction Link**

ziyoung: @akki-jat I mean you can use this example to test in your development environment.

akki-jat: @ziyoung sorry I misunderstood what you said and I also test using that example and its working fine. 
### And you saying put calculating height code in timer?
I don't think it is appropriate to call a code if its not needed.

ziyoung: @akki-jat I tried it again and found that the problem still exists. 

OS/Browsers version: Mac OS X 10_13_6/Chrome/69.0.3497.100

akki-jat: @ziyoung  I don't know what is happening here, is it specific to mac? Because I tested it multiple times on windows 10 / chrome & its working perfectly fine.

akki-jat: @wacky6 / @jikkai can anyone test it? Because it is working fine on my PC. Please check it.

wacky6: Confirm it has issue when `multiple`, `collapse-tags` and `filterable` are set.

See: https://jsfiddle.net/0fj6tpvo/1/
Built against https://github.com/akki-jat/element/commit/b77e59367a8b870f6a6186ffe01a5d9ab802e2c4
Tested on macOS 10.13.6, Chrome 69

akki-jat: @wacky6 / @ziyoung test it again after taking latest.

akki-jat: @ziyoung this problem caused by **resetInputHeight()** function. When we remove a tag this function is call to determine the input height but when this method calculates the height the **( +1 )** tag is not removed ( even if there are no extra selection ). So, when it calculates height it also takes +1 tag into consideration.

Note: UI updates after **resetInputHeight()** function

ziyoung: @akki-jat @ziyoung I optimize `ElTag` component https://github.com/ziyoung/element/blob/dev/packages/tag/src/tag.vue It works fine on my computer.

PR #12410, opened 2018-08-17
Table: Add sort key parameter to sortMethod

johnsusek:
Hello, thank you for creating the great software! This is my first contribution. This change makes custom table sorting more flexible by passing in name of column to sortMethod.

The first two parameters to sortMethod are objects, but the function does not know which key within the object to sort on, so a user has to write different sort functions for every column.

Example of current situation:

```html
<template>
  <el-table :data="tableData">
    <el-table-column
      prop="city"
      label="City"
      sortable
      :sort-method="sortCity">
    </el-table-column>
    <el-table-column
      prop="lastname"
      label="Last Name"
      sortable
      :sort-method="sortLastName">
    </el-table-column>
    <el-table-column
      prop="firstname"
      label="First Name"
      sortable
      :sort-method="sortFirstName">
    </el-table-column>
    <el-table-column
      prop="address"
      label="Address"
      sortable
      :sort-method="sortAddress">
    </el-table-column>
  </el-table>
</template>

<script>
  export default {
    data() {
      return {...}
    },
    methods: {
      sortCity(a, b) {
        if (a.city > b.city) {
          return 1;
        } else if (a.city < b.city) {
          return -1;
        }
        return 0;
        }
    },
      sortFirstName(a, b) {
        if (a.firstname > b.firstname) {
          return 1;
        } else if (a.firstname < b.firstname) {
          return -1;
        }
        return 0;
        }
    },
      sortLastName(a, b) {
        if (a.lastname > b.lastname) {
          return 1;
        } else if (a.lastname < b.lastname) {
          return -1;
        }
        return 0;
        }
    },
      sortAddress(a, b) {
        if (a.address > b.address) {
          return 1;
        } else if (a.address < b.address) {
          return -1;
        }
        return 0;
        }
    },
  }
</script>
```

Right now, user has to create 4 functions - 1 per column.

With proposed change, user can do this with one function.

Example with proposed change:

```html
<template>
  <el-table :data="tableData">
    <el-table-column
      prop="city"
      label="City"
      sortable
      :sort-method="sortAlpha">
    </el-table-column>
    <el-table-column
      prop="lastname"
      label="Last Name"
      sortable
      :sort-method="sortAlpha">
    </el-table-column>
    <el-table-column
      prop="firstname"
      label="First Name"
      sortable
      :sort-method="sortAlpha">
    </el-table-column>
    <el-table-column
      prop="address"
      label="Address"
      sortable
      :sort-method="sortAlpha">
    </el-table-column>
  </el-table>
</template>

<script>
  export default {
    data() {
      return {...}
    },
    methods: {
      sortAlpha(a, b, key) {
        if (a[key] > b[key]) {
          return 1;
        } else if (a[key] < b[key]) {
          return -1;
        }
        return 0;
        }
    },
  }
</script>
```

Thank you!

### DIFF
--- a/examples/docs/en-US/table.md
+++ b/examples/docs/en-US/table.md
@@ -2029,7 +2029,7 @@ You can customize row index in `type=index` columns.
 | fixed | whether column is fixed at left/right. Will be fixed at left if `true` | string/boolean | true/left/right | — |
 | render-header | render function for table header of this column | Function(h, { column, $index }) | — | — |
 | sortable | whether column can be sorted. Remote sorting can be done by setting this attribute to 'custom' and listening to the `sort-change` event of Table | boolean, string | true, false, custom | false |
-| sort-method | sorting method, works when `sortable` is `true`. Should return a number, just like Array.sort | Function(a, b) | — | — |
+| sort-method | sorting method, works when `sortable` is `true`. Should return a number, just like Array.sort | Function(a, b, key) | — | — |
 | sort-by | specify which property to sort by, works when `sortable` is `true` and `sort-method` is `undefined`. If set to an Array, the column will sequentially sort by the next property if the previous one is equal | Function(row, index)/String/Array | — | — |
 | sort-orders | the order of the sorting strategies used when sorting the data, works when `sortable` is `true`. Accepts an array, as the user clicks on the header, the column is sorted in order of the elements in the array | array | the elements in the array need to be one of the following: `ascending`, `descending` and `null` (restores to the original order) | ['ascending', 'descending', null] |
 | resizable | whether column width can be resized, works when `border` of `el-table` is `true` | boolean | — | false |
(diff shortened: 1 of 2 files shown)


jikkai:
Thanks for contributing.

I don't think this is necessary. You can wrap your own `sort-method` like this: https://jsfiddle.net/g4cwd935/

johnsusek:
It is not _absolutely_ necessary, you are correct. But in my opinion, having code like

```
:sort-method="(a, b) => xxx(a, b, 'nnn')"
```

all throughout your source is inelegant - unlike the rest of element-ui which tries to avoid unnecessary code duplication. 

Having this anonymous function defined for every single table column is a lot of unnecessary boilerplate code. And you have to specify 'nnn' in this, creating duplicated strings of column name for no reason. In a large project with a lot of tables, and a lot of columns, this creates a _lot_ of unnecessary code. Also, if you rename columns, it creates a chance for bugs if you forget to rename column name in sort function too.

Frankly, it is ugly. And element-ui should be beautiful.

I feel that:

```
:sort-method="xxx"
```

is just as powerful while still keeping things "as simple as possible" like the rest of element-ui. It also removes chance for bugs if you rename columns, but forget to rename column name in sort function. 

I hope you will reconsider this small unobtrusive change to make element-ui more beautiful, less prone to bugs, but still just as powerful.

 :) ☮️

PR #12411, opened 2018-08-18
Table: Add text search for column filter

johnsusek:
Add text search as option for column filtering in table. 

screen recording 2018-08-17 at 07 11 pm

### DIFF
--- a/examples/docs/en-US/table.md
+++ b/examples/docs/en-US/table.md
@@ -264,6 +264,11 @@
         return row[property] === value;
       },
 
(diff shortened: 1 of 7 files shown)


choudewh:
This is a good solution for text filter.Hope to merge this commit.

akki-jat:
@johnsusek Good solution. But I think there should be two buttons **Apply** and **Reset**.
What do you think?

johnsusek:
@akki-jat I could do that. Curious what @Leopoldthecoder thinks.

PR #12490, opened 2018-08-24
Input: fix success icon overlapping with text

mleg:
When entering text into an input in success status the status icon overlaps with editing area. See image.
image


mleg:
Can't figure out how to rebase. Will make another PR.

PR #12580, opened 2018-09-02
MessageBox: fix hashchange EventListener just Listener once

tanzhen:

The listenchange event in the MessageBox is not deleted in time, causing the page router change to continuously trigger the close function.

There is also a question about message-box using `document.body.appendChild(instance.$el);` when inserting DOM `beforeDestroy` does not trigger, can it be deleted?

### DIFF
--- a/packages/message-box/src/main.vue
+++ b/packages/message-box/src/main.vue
@@ -274,9 +274,13 @@
     },
 
     mounted() {
+      const self = this;
       this.$nextTick(() => {
         if (this.closeOnHashChange) {
-          window.addEventListener('hashchange', this.close);
+          window.addEventListener('hashchange', function close(event) {
+            self.close();
+            event.currentTarget.removeEventListener(event.type, close);
+          });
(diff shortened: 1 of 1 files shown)


packages/message-box/src/main.vue
```diff
@@ -274,9 +274,13 @@
     },
 
     mounted() {
+      const self = this;
```
ziyoung: 或者你可以直接使用这个 once 函数。
```js
import { once } from 'element-ui/src/utils/dom';
```

packages/message-box/src/main.vue
```diff
@@ -274,9 +274,13 @@
...
       this.$nextTick(() => {
         if (this.closeOnHashChange) {
-          window.addEventListener('hashchange', this.close);
+          window.addEventListener('hashchange', function close(event) {
```
wacky6: 这里直接用 arrow function，不需在外面加 self

Issue #12732, opened 2018-09-12
[Bug Report] Form-Item does not emit individual errors as they happen


YamenSharaf:
### Element UI version
2.4.6

### OS/Browsers version
All

### Vue version
2.5.17

### Reproduction Link
[https://jsfiddle.net/mkv54a6s/](https://jsfiddle.net/mkv54a6s/)

### Steps to reproduce
Try to react to errors on individual form items by displaying a different kind of error messages

### What is Expected?
To be able to react to errors and have my own custom validation strategy. have more general flexibility

### What is actually happening?
Not possible to react to errors in a different way than existing implementation

YamenSharaf:
Created PR to fix this issue by emitting an array of errors when an error occurs. User can then listen to that event which provides the name of the field associated and the error message.

https://github.com/ElemeFE/element/pull/12733

ziyoung:
`Validate` is added in pr #10351. I think you also can get error info by listening the `validate` event.

YamenSharaf:
Validate event works on the form as a whole, so it's not easy to make a component to wrap each form element and handle them individually, which is what I'm trying to rectify with this.

I made this PR and [another one](https://github.com/ElemeFE/element/pull/12715) to essentially make errors more customizable by the end user since currently, it's unwieldy to try to change how error messages look like in a different way than the default behaviour (red text under form element).

YamenSharaf:
One more thing. The Form element's validate event does not emit the error message (Which should be fixed too imo). It just returns the validation status as a boolean which isn't very useful.

ziyoung:
ref #12860

PR #12733, opened 2018-09-12
FormItem: Form item emits array of errors when an error occurs

YamenSharaf:


YamenSharaf:
This is to fix this issue I posted [here](https://github.com/ElemeFE/element/issues/12732) earlier

https://github.com/ElemeFE/element/issues/12732

YamenSharaf:
@ziyoung This is the payload emitted by the Form component at the moment. It does not emit the current error message.

screen shot 2018-09-12 at 10 04 14 pm

Personally, I think both Form and FormItem should emit that error message to enable developers to wrap these components in any way they want.

 I disagree with how the payload is being evaluated right now on the Form (as an array). An object of declared keys would be the way to go usually, but for the sake of backwards compatibility, I think It should be added as the third entry in that array anyway.

abdallanayer:
@ziyoung I agree with @YamenSharaf that we need more flexibility with the error messages to give the developers more power to control how they show the errors

ibufu:
@YamenSharaf Is it better to put the error message in the validate event of form?

ziyoung:
If validate event emits error message, is it necessary to add error event?

YamenSharaf:
@ibufu @ziyoung Personally I think we should have both, because:

- The bigger use case for this is listening to a specific error on a specific form-item. Probably to take action on such error.

- Each form item could have multiple errors associated with it at the time (Although it's showing one at a time). having a list of errors would be useful to trigger an action on specific error

- The current validate event on form should only care about the errors visible to the end-user, and the way its payload is structured is not ideal (but it makes sense for it to emit the error itself anyway).

YamenSharaf:
@ziyoung Also made [this PR](https://github.com/ElemeFE/element/pull/12860) for having form emit the actual error on validate events. This doesn't seem very wieldy since ElForm emits that event all the time for each individual form element. I still think it can be useful though. https://github.com/ElemeFE/element/pull/12860

wacky6:
`errors` are actually a list of `new Error(message)`. These messages are meant to be read by user (not to be used in program logic).

What would you plan to do with these strings? For example?

```
if (errors.find(e => e.message.includes('Field A is required'))) {
    blame_user_with_dialog(...)
}
```

String comparison (on error message) does not make sense to me.

YamenSharaf:
@wacky6 String comparison is icky, but sometimes you wanna respond to a certain error in a certain way, let that be dimming a button or popping an alert. It could be that I want errors to appear on the side of the form as opposed underneath inputs.

The way things are set up now there's no way to access these errors for any reason which I don't think is very flexible.

wacky6:
In that case, I would append additional codes after calling validate(), and check the v-model / form's value directly.

Validators are supposed to show a message, but nothing more.

Issue #12713, opened 2018-09-11
[Bug Report] About the echo of el-select under multiple, remote, and value-key conditions! Urgent!


yang123guo:
### Element UI version
2.4.6

### OS/Browsers version
无关操作系统和浏览器版本

### Vue version
2.5.16

### Reproduction Link
[https://jsfiddle.net/1erjhcL3/773/](https://jsfiddle.net/1erjhcL3/773/)

### Steps to reproduce
回显内容 看不见，
只有再次change以后才会出现
![www](https://user-images.githubusercontent.com/17383237/45364303-5ebb5c00-b60c-11e8-9828-3f930723bbf0.png)

### What is Expected?
回显能不触发远程调用，把内容显示出来

### What is actually happening?
回显内容 看不见，只有再次change以后才会出现

yang123guo:
@nanomen   Thanks for your suggestion, I"ll try it!

nanomen:
@yang123guo you're welcome! If you want, look the fixed fork in my repo https://github.com/nanomen/element

And you can write own select element, just use root select element as extend in your own select element.

```js
// own select

import { Select } from 'element-ui'

export default {
  name: 'OwnSelect',
  extend: Select,
  props: {
    labelKey: {
       type: String,
       default: 'label'
    }
  },
  methods: {
    getOptions (value) {
      ...
      const label = (isObject && !isNull) 
        ? value[this.labelKey] ? value[this.labelKey] : ''
        : (!isObject && !isNull) ? value : '';
      ...
    }
  }
}
```

and use in your code.

yang123guo:
got it! @nanomen

ziyoung:
https://github.com/ElemeFE/element/pull/12735#issuecomment-422088645

PR #12735, opened 2018-09-12
select: add label-key property (#12713)

nanomen:
Fix https://github.com/ElemeFE/element/issues/12713

Add in el-select a new property "label-key" which the same as value-key. This allow to set a label for tag element when we pass value as Object to select component.

### DIFF
--- a/examples/docs/en-US/select.md
+++ b/examples/docs/en-US/select.md
@@ -665,6 +665,7 @@ If the binding value of Select is an object, make sure to assign `value-key` as 
 | multiple | whether multiple-select is activated | boolean | — | false |
 | disabled | whether Select is disabled | boolean | — | false |
 | value-key | unique identity key name for value, required when value is an object | string | — | value |
+| label-key | unique identity key name for label, required when value is an object | string | — | label |
 | size | size of Input | string | large/small/mini | — |
 | clearable | whether single select can be cleared | boolean | — | false |
 | collapse-tags | whether to collapse tags to a text when multiple selecting | boolean | — | false |
--- a/examples/docs/en-US/select.md
+++ b/examples/docs/en-US/select.md
@@ -665,7 +665,7 @@ If the binding value of Select is an object, make sure to assign `value-key` as 
 | multiple | whether multiple-select is activated | boolean | — | false |
 | disabled | whether Select is disabled | boolean | — | false |
 | value-key | unique identity key name for value, required when value is an object | string | — | value |
-| label-key | unique identity key name for label, required when value is an object | string | — | label |
+| label-key | unique identity key name for label. Use only when value is an object and the label doesn't display | string | — | label |
 | size | size of Input | string | large/small/mini | — |
 | clearable | whether single select can be cleared | boolean | — | false |
 | collapse-tags | whether to collapse tags to a text when multiple selecting | boolean | — | false |
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -303,6 +303,10 @@
         type: String,
         default: 'value'
       },
+      labelKey: {
+        type: String,
+        default: 'label'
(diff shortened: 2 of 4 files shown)


nanomen:
Hello! I don't know what's happend (about test error "fullscreen singleton"). In my local machine all tests is passed image

wacky6:
Why is this necessary?

nanomen:
@wacky6 If I using `label` as property of object it is empty. Issue [#12713](https://github.com/ElemeFE/element/issues/12713) is fully describe the bug. Look, please.

ziyoung:
@nanomen Could you please update doc for this attribute? I think you can also add a tip for this.

nanomen:
@ziyoung sure! one moment!

nanomen:
@ziyoung done! Add in this page `/examples/docs/en-US/select.md`

examples/docs/en-US/select.md
```diff
@@ -665,6 +665,7 @@ If the binding value of Select is an object, make sure to assign `value-key` as
 | multiple | whether multiple-select is activated | boolean | — | false |
 | disabled | whether Select is disabled | boolean | — | false |
 | value-key | unique identity key name for value, required when value is an object | string | — | value |
+| label-key | unique identity key name for label, required when value is an object | string | — | label |
```
ziyoung: - I think this attribute should not be required. Only when the label doesn't display, the attribute is needed. You can add a tip for this.
- Just copy these change to Chinese and Spanish docs.

packages/select/src/select.vue
```diff
@@ -533,8 +537,9 @@
           }
         }
         if (option) return option;
-        const label = (!isObject && !isNull)
-          ? value : '';
+        const label = (isObject && !isNull)
+          ? value[this.labelKey] ? value[this.labelKey] : ''
```
ziyoung:
```js
const label = (isObject && !isNull)
  ? (value[this.labelKey] || '')
  : (!isObject && !isNull) ? value : '';
```

nanomen:
@ziyoung add description in docs and refactor conditional statement, thank you!

wacky6:
My opinion is **strongly against** this feature. I have seen a similar issue some time ago. From my perspective, `label-key` is permitting developer sloppiness and laziness, which should never be tolerated in the first place.

A select without el-option should show nothing, even values are specified. It's *developer's* responsibility to add selected option into el-option list.

This is **analogous to native HTML select**. See https://jsfiddle.net/dphut543/, native select shows empty value if bounded value is not in the option list.

The correct way of fixing #12713 is https://jsfiddle.net/p0h327dw/. Basically, ensure the selected value is contained in option list.

---

Object values are meant to enable representation of complex values (for example, configuration preset / combination of values). They are not supposed to contain the label, which is rarely required for application logic (or submitted to the server).

The reason that `value-key` was implemented is to determine equality of different objects. Another way is ask developer to write a comparison function (like the one in `Array.sort`), which is laborious (and usually not required).

wacky6:
@nanomen 

By saying "laziness", I mean developers omitting the step of "ensuring selected options are contained in the list of available options, and think about what labels should be displayed". In other words, logically correct (Some might consider this pedantic). This step could be as simple as merging two lists together.

I understand that it is tempting to show object values in some way (i.e. label-key). But it does not make sense logically. Suppose bounded values are integers, and options are empty. Current implementation displays the bounded value. (https://jsfiddle.net/3xyopdf0/).

This is acceptable, but logically incorrect. Select means choosing a subset of available options:

* Why would someone show internal values instead of user-friendly labels to user?
* Why would select display an option that is not permitted (declared by el-option)?
* How could someone select an option when there none is available?

---

In fact, I would rather el-select ignores all values that are not presented in option list. The current behavior is due to legacy implementation.

PR #12801, opened 2018-09-18
form: replace reject(valid) with resolve(valid)

callmewhy:
For now, if the "valid" is false, it will return a `reject(valid)` in `await` and we must wrap the validate call with a `try catch`:

```js
let result = false
try {
  result = await this.$refs.investForm.validate()
} catch (error) {
  result = false
}
if (result) {
// xxx
}
```

If we use resolve(valid), it will be very suitable to handle the validator result:

```js
const result = await this.$refs.form.validate()
if (result) {
// xxx
}
```

"validator result" is a result value, not an exception.
I think `valid === false` is an expected result, it's better to resolve(valid) instead of reject(valid).

### DIFF
--- a/packages/form/src/form.vue
+++ b/packages/form/src/form.vue
@@ -102,7 +102,7 @@
         if (typeof callback !== 'function' && window.Promise) {
           promise = new window.Promise((resolve, reject) => {
             callback = function(valid) {
-              valid ? resolve(valid) : reject(valid);
+              resolve(valid);
             };
           });
         }


wacky6:
This won't be merged for now. It is breaking change.

try `await this.$refs.investForm.validate().then(ok => true, error => false)`

wacky6:
If resolve, then there is no way to get which validation fails.

Also, Promise.then already offers a way to branch application logic based on valid/invalid.

callmewhy:
> If resolve, then there is no way to get which validation fails.
A: for now, with `reject(false)`, we cannot get which validation fails too. we will only catch an error: false.

> Also, Promise.then already offers a way to branch application logic based on valid/invalid.
if invalid, we have to `try catch` the exception. 

It's not an error, it's a valid result. We shouldn't use `throw error` to handle valid result. `false/true` is an expected result.

kefranabg:
Please consider merging this, the validate should never throw an error when the form is invalid as this is not a critical error, it just the result of the result of the validate method.

CarterLi:
I never try/catch those errors. Instead I set up a global error handler.

```ts
const errorHandler = event => {
  if (!event.reason) return;

  if (event.reason.isAxiosError) {
    $message.fuck(
      event.reason,
      !(event.reason instanceof Error),
    );
    return;
  }

  if (event.reason instanceof Error) {
    if (event.reason.name === 'NavigationDuplicated') {
      return;
    }

    $message.fuck(event.reason);
    return;
  }
};

window.addEventListener('unhandledrejection', errorHandler);

Vue.config.errorHandler = reason => {
  errorHandler(reason);
};
```

When using form, I just write

```js
await this.$refs.form.validate();
await XxxApi.save(this.model);
this.$message.success('Success');
```

No `try/catch`, no `if`s

callmewhy:
> I never try/catch those errors. Instead I set up a global error handler.

It doesn't make sense if I want to show an error message below the input.

With this solution, we try to change our UX because of the UI framework. I think it's a workaround for personal project, but it's not a solution.

CarterLi:
> > I never try/catch those errors. Instead I set up a global error handler.
> 
> It doesn't make sense if I want to show an error message below the input.

Isnt it the default behavior of ElForm#validate?

> 
> In this way, we try to change our UX because of the UI framework. I think it's a workaround for personal project instead of a solution.

We use it in our production code, and it always works fine

callmewhy:
> We use it in our production code, and it always works fine

You can catch all exception and handle them in `errorHandler`, it's totally ok. It works fine for you but it doesn't works for the guys who want to handle the exception in current page instead of a global callback.

"show an error message below the input" is just an example to show that when you choose to catch form validation result with `unhandledrejection`, you give up the ability to customize the UX in current page. 


My point is that, we shouldn't throw an exception when the result is expected. It's not a problem of can or cannot but a discussion about better design.

CarterLi:
> > We use it in our production code, and it always works fine
> 
> You can catch all exception and handle them in `errorHandler`, it's totally ok. It works fine for you but it doesn't works for the guys who want to handle the exception in current page instead of a global callback.
> 
> "show an error message below the input" is just an example to show that when you choose to catch form validation result with `unhandledrejection`, you give up the ability to customize the UX in current page. And I don't think it's a good practice to rely on `unhandledrejection ` because there will be lots of switch case to handle different kinds of errors which should be handled in the lib or pages.

If you want to handle it in current page, you can still use try/catch.

The question is that why do you want to handle the exception of ElForm#validate? The *default behavior* of ElForm#validate already shows error messages below the input box. I dont think there are other things you can do.

> 
> My point is that, we shouldn't throw an exception when the result is expected. It's not a problem of can or cannot but a discussion about better design.

Issue #2037, opened 2016-12-28
RTL support


mahmoudmy:
Thanks guys, 
That's a really nice project.
Do you have any plan for *RTL* support in css? (for languages like Persian, Arabic, ...)

Leopoldthecoder:
Thanks. 
We don't have plans to add RTL support, but there may be some third party libraries you can use with Element, like [rtlcss](https://github.com/MohammadYounes/rtlcss).

MahdiMajidzadeh:
no plan to add rtl with contribution neither ??

osman-mohamad:
do you have any guides about using element-ui with rtlcss ?

KayvanMazaheri:
Has anyone find a solution for using _Element_ with RTL languages? Any suggestions would be appreciated.
@mahmoudmy @MahdiMajidzadeh @osman-mohamad

osman-mohamad:
@KayvanMazaheri I did found a solution . you can use postcss-loader in your project with postcss `rtlcss` plugin . and it will convert all the styles to rtl .
plus :
```css
body,html {
   direction: ltr
}
``` 
in the App.vue Component style

siavash-sajjad:
@osman-mohamad 
hi my friend
can you help me to add and usage rtlclass to laravel-vue project?

osman-mohamad:
@siavash-sajjad put the repository url here . for me to review it .

kooshkestani:
@Leopoldthecoder hi tree element not support rtl .
how to customize  tree-node.vue 
`      :style="{ 'padding-left': (node.level - 1) * tree.indent + 'px' }">
`to padding-right???

moosti:
+1

ghyath5:
How can I use rtlcss in my nuxtjs project ?
and did you find another way to support RTL in element?

kooshkestani:
https://www.npmjs.com/package/tree-element-ui-rtl

saeed-darestany:
+1

hitautodestruct:
+1

realSaeedHassani:
@Leopoldthecoder  Still have no plans for RTL support?

mra-dev:
It would be so good if had it.

mahsamirzapour:
> Thanks guys, That's a really nice project. Do you have any plan for _RTL_ support in css? (for languages like Persian, Arabic, ...)

Hi homie, just use like this:
html {
direction: rtl;
}
[class*='el-col-'] {
[...]

PR #12858, opened 2018-09-21
RTL Support

Mont4:
Add rtl support to Element package. #2037


ziyoung:
Element is a ui library. It can't satisfy the needs of all users. RTL is a feature but not the most wanted feature. Where to use RTL CSS should be determined by the user.

PR #12961, opened 2018-10-02
[Docs] Fix info about Input events

dodas:
- Document Input's "input" event, similar to HTML's input element
- Fix mistake in Input's "change" event description

### DIFF
--- a/examples/docs/en-US/input.md
+++ b/examples/docs/en-US/input.md
@@ -682,7 +682,8 @@ Search data from server-side.
 |----| ----| ----|
 | blur | triggers when Input blurs | (event: Event) |
 | focus | triggers when Input focuses | (event: Event) |
-| change | triggers when the icon inside Input value change | (value: string \| number) |
+| change | triggers when the Input value changes and loses focus | (value: string \| number) |
+| input | triggers when the Input value changes | (value: string \| number) |
 | clear | triggers when the Input is cleared by clicking the clear button | — |
 
 ### Input Methods


examples/docs/en-US/input.md
```diff
@@ -682,7 +682,8 @@ Search data from server-side.
 |----| ----| ----|
 | blur | triggers when Input blurs | (event: Event) |
 | focus | triggers when Input focuses | (event: Event) |
-| change | triggers when the icon inside Input value change | (value: string \| number) |
+| change | triggers when the Input value changes and loses focus | (value: string \| number) |
+| input | triggers when the Input value changes | (value: string \| number) |
```
wacky6: Input event is implied by Vue. There is no need to include it in doc.

dodas: Ah, I forgot it's part of `v-model` feature..do you think it's not worth any mention?

Other than that, description of `change` needs to be fixed:
>triggers when the icon inside Input value change
what icon? :D

wacky6: Element is a Vue framework, we expect user to be competent in Vue before using Element. The change's description indeed requires fixing. :D

dodas: I am saying that Docs doesn't mention `v-model` either, so that needs to be assumed too :D 

Regarding `change` event: should we mention the difference between `input` and `change` (that it triggers after blurring), or just fix `icon` error?

wacky6: Yes. v-model is also assumed knowledge. It is covered extensively in vue's documentation.

Just fix the error.  `change` event generally behaves similarly to native component's change event.

ziyoung: There's no need to add `input` event.

ziyoung:
@dodas update Chinese and Spanish docs.

PR #12972, opened 2018-10-05
Tabs: use inject for all internal components

vincenting:

* [√ ] Make sure you follow Element's contributing guide ([中文](https://github.com/ElemeFE/element/blob/master/.github/CONTRIBUTING.zh-CN.md) | [English](https://github.com/ElemeFE/element/blob/master/.github/CONTRIBUTING.en-US.md) | [Español](https://github.com/ElemeFE/element/blob/master/.github/CONTRIBUTING.es.md)).
* [√ ] Make sure you are merging your commits to `dev` branch.
* [ √] Add some descriptions and refer relative issues for you PR.

Reactor component tabs, remove values passing by props and use inject instead.
Since now we use both props and inject in this component.

### DIFF
--- a/packages/tabs/src/tab-bar.vue
+++ b/packages/tabs/src/tab-bar.vue
@@ -5,13 +5,12 @@
   export default {
     name: 'TabBar',
 
-    props: {
-      tabs: Array
-    },
-
     inject: ['rootTabs'],
 
     computed: {
+      tabs() {
+        return this.rootTabs.panes;
+      },
       barStyle: {
         cache: false,
         get() {
--- a/packages/tabs/src/tab-nav.vue
+++ b/packages/tabs/src/tab-nav.vue
@@ -2,10 +2,29 @@
   import TabBar from './tab-bar';
   import { addResizeListener, removeResizeListener } from 'element-ui/src/utils/resize-event';
 
-  function noop() {}
   const firstUpperCase = str => {
     return str.toLowerCase().replace(/( |^)[a-z]/g, (L) => L.toUpperCase());
   };
+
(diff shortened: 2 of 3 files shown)


ziyoung:
Does this look more concise? By passing data to child components with props, we can know attributes of a component and specify types of attributes.

wacky6:
Wouldn't it be better to use props instead of inject?

If everything is done in props, tab nav can be used in a standalone way (though some hack is required)

vincenting:
Hi, when we a design or implement a component, I think we need to define the use cases for each component. That means there always has some public and private component in our project. And I also I agree it will be great if we can make all components reusable.

**Inject/Provider is only an ability to help us to reduce passing value through multi-level components, or we want to maintain all of our data in the root component for an atomic component**. Tabs component have a structure like Tabs -> TabNav -> TabBar. It is reasonable to use inject to reduce the nested value passing by props (In Tab Bar). Otherwise, let's imagine we not only want to get a value from the Tabs but also want to change the state in Tabs. It will be much more complex. Inject/Provider is just like the component instance level Vuex from the abstract. In my current working company, we have a private library to maintain the provider state in the same way as Vuex.

Back to this discussion, if we want to reuse tab-nav as a standalone component, we might need to design another component if we also want to provide a component which only contains nav-bar, and emits the event when the user selects a nav item without the integration of tab-pane. That makes NavBar another atomic component.

**I think that makes sense. And it is not some kind of hack I think but refactoring work is actually needed. In that design, tabs only need to pass the value to tabs bar component and receive the callback (event) from nav-bar, so inject is not needed then**.

PR #12990, opened 2018-10-09
Form: fix string type param bug

codinglobster:

here's the explain, when we pass a string type param to `form.clearValidate` method it also will work, but in some situation it perform weird, e.g. if we have two `form-item`, their props contain same keyword, like` 'password'` and `'re-password'`, then if we use `form.clearValidate('re-password')`,both of them's status will be clear, but I only want to clear `re-password`.
I know the doc tell us to pass a array type param, but I think string type param should be supported.

### DIFF
--- a/packages/form/src/form.vue
+++ b/packages/form/src/form.vue
@@ -85,7 +85,10 @@
       },
       clearValidate(props = []) {
         const fields = props.length
-          ? this.fields.filter(field => props.indexOf(field.prop) > -1)
+          ? (typeof props === 'string' 
+              ? this.fields.filter(field => props === field.prop)
+              : this.fields.filter(field => props.indexOf(field.prop) > -1)
+            )
           : this.fields;
         fields.forEach(field => {
           field.clearValidate();


packages/form/src/form.vue
```diff
@@ -85,7 +85,10 @@
       },
       clearValidate(props = []) {
         const fields = props.length
```
ziyoung: The type of `props` is expected to be an array. When `props` is a string, `props.length` maybe truthy. I think there's no need to support string param.

wacky6: Accepting a string won't hurt.

Given the source code, from user's perspective: 

* if props is not given or an empty array: clear all fields
* if props is given, and is a non-empty array, clear fields in the array

---

In my opinion, the API of this function should be fixed:

* if props is not given (undefined / null): clear all fields
* if props is a string: clear the field indicated by props
* if props is an array: clear fields indicated by strings in the array

Clear all fields when given an empty array does not make sense to me.

Issue #11807, opened 2018-06-29
[Feature Request] Can add  table columns filter slot at el-table?


ruifeng96150:
### Existing Component
Yes

### Component Name
el-table

### Description
Sometimes, users want to search some thing at special column, so I think it is possible to implement at el-table? Yes, the requirement that column with enumeration filter has been implemented at current version. But some columns may need date range or number range or input keyword and so on. So slot and function interface may allow user to define their filters. This slot can placed between the table header and table data.

verzilka:
+1 I need this too.
I had to create my own wrapper instead of filter-panel and patch table elements in ugly way =(
Hope this can have native support. Please )

iuxw:
+1

sotaan:
+1 plz this will be usefull

Mont4:
Similar to https://onewaytech.github.io/vue2-datatable/examples/dist/

<img width="1440" alt="screen shot 2018-12-29 at 9 17 38 pm" src="https://user-images.githubusercontent.com/41866067/50540866-39e4f780-0baf-11e9-8c1a-d3c28510e118.png">

Mortimer2013:
+1

PR #13012, opened 2018-10-10
Table: Adds header rendering slot

### LINKED ISSUE #4908: [Question] Custom header for the el-table-column
Hello!

Could i have a custom template for the column header? 

I need to have some filtering using el-input.

Endresult should be similar to

screenshot_2017-05-17_13-13-53

### LINKED ISSUE #4909: [Feature Request] Custom header for the el-table-column
### Existing Component
Yes

### Component Name
el-table-column

### Description
Hello!

Could i have a custom template for the column header?

I need to have some filtering using el-input.

Endresult should be similar to

screenshot_2017-05-17_13-13-53

ivanseidel:
Fixes #4909 
Fixes #4908
Related #11807
Related #6573 

Now it is possible to assign custom headers directly to templates, exactly as renderHeader does.

### DIFF
--- a/examples/docs/en-US/table.md
+++ b/examples/docs/en-US/table.md
@@ -2125,3 +2125,4 @@ You can customize row index in `type=index` columns.
 | Name | Description |
 |------|--------|
 | — | Custom content for table columns. The scope parameter is { row, column, $index } |
+| header | Custom content for table header. The scope parameter is { column, $index } |
--- a/examples/docs/en-US/table.md
+++ b/examples/docs/en-US/table.md
@@ -185,6 +185,23 @@
           amount2: '4.1',
           amount3: 15
         }],
+        tableData7: [{
+          date: '2016-05-02',
+          name: 'Tom',
+          address: 'No. 189, Grove St, Los Angeles',
+        }, {
+          date: '2016-05-04',
+          name: 'John',
+          address: 'No. 189, Grove St, Los Angeles',
+        }, {
+          date: '2016-05-01',
+          name: 'Morgan',
+          address: 'No. 189, Grove St, Los Angeles',
+        }, {
+          date: '2016-05-03',
+          name: 'Jessy',
+          address: 'No. 189, Grove St, Los Angeles',
+        }],
         currentRow: null,
         multipleSelection: [],
         search: '',
@@ -1508,7 +1525,7 @@ Customize table header so it can be even more customized.
 ```html
 <template>
   <el-table
-    :data="tableData.filter(data => !search || data.name.toLowerCase().includes(search.toLowerCase()))"
+    :data="tableData7.filter(data => !search || data.name.toLowerCase().includes(search.toLowerCase()))"
     style="width: 100%">
     <el-table-column
       label="Date"
@@ -1556,18 +1573,23 @@ Customize table header so it can be even more customized.
           address: 'No. 189, Grove St, Los Angeles'
         }, {
           date: '2016-05-02',
-          name: 'Tom',
+          name: 'John',
           address: 'No. 189, Grove St, Los Angeles'
         }, {
           date: '2016-05-04',
-          name: 'Tom',
+          name: 'Morgan',
           address: 'No. 189, Grove St, Los Angeles'
         }, {
           date: '2016-05-01',
-          name: 'Tom',
+          name: 'Jessy',
           address: 'No. 189, Grove St, Los Angeles'
-        }]
+        }],
+        search: '',
       }
+    },
+    methods: {
+      handleEdit(){},
+      handleDelete(){}
     },
   }
 </script>
--- a/examples/docs/en-US/table.md
+++ b/examples/docs/en-US/table.md
@@ -1536,13 +1536,6 @@ Customize table header so it can be even more customized.
       prop="name">
     </el-table-column>
     <el-table-column
-      v-for="header in extraColumns"
-      :key="header"
-      :label="header"
-      prop="name">
-    </el-table-column>
-    <el-table-column
-      label="Name"
       align="right">
       <template slot="header" slot-scope="slot">
         <el-input
--- a/examples/docs/en-US/table.md
+++ b/examples/docs/en-US/table.md
@@ -186,7 +186,8 @@
           amount3: 15
         }],
         currentRow: null,
-        multipleSelection: []
+        multipleSelection: [],
+        search: '',
       };
     },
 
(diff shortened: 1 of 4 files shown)


ziyoung:
TableColumns has `render-header` attribute which allows you custom column content. Is it necessary to add header named slot? If I set `render-header` and use header slot, how to render this column?

ivanseidel:
Slots are more easy to create than render function, as you dont need to write any sort of component code, just template.

See how hard it is to write custom render functions for pretty simple things in #4909.

My initial idea was to create a scope and to create another check in table-header to render it, but then I noticed that the syntax of render-header could be used in the same way.

I would say that scoped slots should take precedence, just like a render-header takes precedence on simple label rendering. It isnt that different, as current behavior opts for only one of the methods.

As you can see in my code, this change doesnt adds much overhead, since the template is only being aplied to the renderHeader prop during mounting. The rest is just like it was supposed to be: a custom renderHeader function.

Also, if we agree that "renderHeader" should take precedence over slots (witch I think is not good in my opinion), is just a matter of checking for the prop and avoiding to set it. It can be possible in both ways...

ziyoung:
@ivanseidel you can use vue-jsx to set `render-function`. It's not hard to create than render function.
https://codesandbox.io/s/w2jyqwjzql
 If there are two ways to custom column content, this will cause confusion.

ivanseidel:
I dont think it will cause confusion. This should be discussed.

I spent over two hours to work with a render function in my controller, and I think I spent my time for nothing as I could easily have done that just like the docs says on the body funcion of the cell.

It doesnt make sense to let users decide their cell content on scoped slots but not the header. If that made sense it should be both render function or scoped slots, not one different from the other.

From the perspective of simplicity, it is much more easy to work with slots, also, it is The documented way on Vue docunentation to do so. Assigning functions as parameters to be called a single time is not a common scenario on vue, and should be a anti-pattern.

I suggest to get more people envolved, because I love Element and I see this is an easy inprovement with no conflicts that will make people life much easier to create custom tables.

Thankz

ivanseidel:
Also, the main idea of separating Vue.js controller, style and template is to avoid having to work on layout, styling and animations in Controllers. 

Your example lacks on showing this working, since there is a method responsible for rendering elements. It should be avoided, to make code cleaner and more maintainable. This looks more like `React` than `Vuejs`. Maybe it is a good approach for `React`.

``
`
methods: {
    renderHeader(h, { column }) {
      return <el-button type="primary">{ column.label }</el-button>
    }
  }
```

**Comment by ziyoung** [comment]
@ivanseidel Yes. Compared to render function, slots are easier to use.
If we add the header named slot, there are two ways to custom table column header.
I'm reminded of this sentence in [The Zen of Python](https://www.python.org/dev/peps/pep-0020/).
> There should be one-- and preferably only one --obvious way to do it.

So, if header slot is supported, should we deprecate `render-header` function in next major version and recommend users to use header slot?

**Comment by ivanseidel** [comment]
For me we have 3 options to choose:

1. Add slots, and make it "priority" over renderHeader.

2. Add slots and make it "less priority" (will opt for renderHeader instead of slot)

3. Go with option 1, and log a warning of deprecation if user inputs anything in parameter (justlike you said).

I would go for option 3, as we would be able to "measure" how many people still require thia feature by showing the link of thiz pull request. If people really need renderHeader after all, we could just make it something ocluded from the docs, or less direct for users. They would then use primarilly slots and if they need, renderHeader.

What so you think?

**Comment by ziyoung** [comment]
@ivanseidel I agree with you.

File: examples/docs/en-US/table.md [hunk 1]
```diff
@@ -1500,6 +1501,79 @@ Customize table column so it can be integrated with other components.
 ```
 :::
 
+### Table with custom header
```
ziyoung: Update the missing Spanish and Chinese docs (You don't need to translate).

ivanseidel: I did manage to add Spanish, but couldn't find the right spot on Chinese, can you give me a hand on that?

ivanseidel:
Hi @ziyoung can you check those changes?

examples/docs/en-US/table.md
ziyoung: https://github.com/ElemeFE/element/blame/dev/examples/docs/zh-CN/table.md#L1540

packages/table/src/table-column.js
```diff
@@ -294,6 +294,11 @@ export default {
       }
     });
 
+    // Deprecation warning for renderHeader property
+    if (this.renderHeader) {
+      console.warn('[Element Warn][Table Column]\'render-header\' property will be deprecated in next major version. please use scoped-slot \'header\' instead.');
```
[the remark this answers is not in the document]
ziyoung: Maybe we can't deprecate it. I find `render-header` is widely used.😂

ziyoung:
You also need to add `Table-column Slot` at the bottom of the document page.

ivanseidel:
Done @ziyoung! All done I guess

packages/table/src/table-column.js
```diff
@@ -444,6 +449,10 @@ export default {
       columnIndex = [].indexOf.call(parent.$el.children, this.$el);
     }
 
+    if (this.$scopedSlots.header) {
```
ziyoung: For selection column, header scope-slot is not allowed to set. When users use this slot, they should get warning information.

packages/table/src/table-column.js
ziyoung: I find `render-header` is widely used. Maybe we can't deprecate this attribute in next major version. We should give users enough time to upgrade. 
A potential warning info is like this:
> Comparing to `render-header`, scoped-slot header is easier to use. We recommend users to use scoped-slot header.

examples/docs/en-US/table.md
```diff
@@ -1500,6 +1501,79 @@ Customize table column so it can be integrated with other components.
...
+    style="width: 100%">
+    <el-table-column
+      label="Date"
+      prop="date">
+    </el-table-column>
+    <el-table-column
+      label="Name"
+      prop="name">
+    </el-table-column>
+    <el-table-column
```
ziyoung: I find this column is redundant.

examples/docs/en-US/table.md
```diff
@@ -1500,6 +1501,79 @@ Customize table column so it can be integrated with other components.
...
+          @click="handleDelete(scope.$index, scope.row)">Delete</el-button>
+      </template>
+    </el-table-column>
+  </el-table>
+</template>
+
+<script>
+  export default {
+    data() {
+      return {
```
ziyoung: `search` should be included in the return value.

examples/docs/en-US/table.md
```diff
@@ -1500,6 +1501,79 @@ Customize table column so it can be integrated with other components.
...
+          date: '2016-05-04',
+          name: 'Tom',
+          address: 'No. 189, Grove St, Los Angeles'
+        }, {
+          date: '2016-05-01',
+          name: 'Tom',
+          address: 'No. 189, Grove St, Los Angeles'
+        }]
+      }
+    },
```
ziyoung: You should add missing methods.
```js
methods: {
 handleEdit(){},
 handleDelete(){}
}
```

ivanseidel:
Should be ok now @ziyoung

ivanseidel:
Should be ok now @ziyoung

ivanseidel:
😍😍😍

KnorienChang:
> This is not working in 2.4.9. does anyone could help me ?
> 
> The guide said: Custom content for table header. The scope parameter is { column, $index }
> 
> my code:
> 
> ```
> <el-table-column prop="name" label="Name">
>         <template slot="header" slot-scope="scope">
>           <el-input @input="test(scope)" size="mini" placeholder="Type to search" />
>         </template>
>       </el-table-column>
> ```
> while here scope is type of function, am i do something wrong?

slot-scope="slot"
example:https://github.com/ElemeFE/element/pull/13012/files/991ef98f67d5380c26af6d9bf09c1f7dbda5c7e1#diff-f9a0f43892aff2d5d46d321fa34b903cR1530

dhthomas76:
The documentation for the new header scoped-slot says "Custom content for table header. The scope parameter is { column, $index }", but I am unable to access the column or $index properties.  Whenever I access the "slot" variable it simply returns a render function (not an object with the column and $index properties).  Could someone provide an example of using the new header scoped-slot while accessing the column within the template? 

`<template slot="header" slot-scope="slot">
		       <div>{{slot.column.label}}</div>
</template>`

felipe-augusto:
@Flex-Xuan @KnorienChang @dhthomas76 This issue has been fixed on PR #13263, but has not been released yet.

Flex-Xuan:
> @Flex-Xuan @KnorienChang @dhthomas76 This issue has been fixed on PR #13263, but has not been released yet.

Got it, thanks

yolofit:
#21320

PR #13042, opened 2018-10-12
InputNumber: event on typed changes

goatandsheep:
When people are typing into number input, they need to receive change notification

### DIFF
--- a/packages/input-number/src/input-number.vue
+++ b/packages/input-number/src/input-number.vue
@@ -39,6 +39,7 @@
       @keydown.down.native.prevent="decrease"
       @blur="handleBlur"
       @focus="handleFocus"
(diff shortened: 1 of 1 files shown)


wacky6:
Not handling 'input' event is intentional. This is to allow partial inputs such as '0.', which is emitted as 'input' event. If handleInputChange is called, such inputs will be truncated, and user will not able to input decimal numbers.

goatandsheep:
I think it can work if we update on input and format on change

wacky6:
Well. It won't. Input-number's input emits a number. There is no way to distinguish "0." and "0" which are both represented as Number(0).

When input is emitted, component's value gets updated, which resets the value of native input.

PR #13104, opened 2018-10-21
Fix popover issues with hover display.

goldengecko:
There are two issues addressed in this pull request:

1. When the trigger is set to "hover", the popover is also triggered on focus, which is incorrect behaviour. Fixed by setting the handling of handleFocus and handleBlur to just show/hide the popover when the trigger is set to "click" or "focus".

2. When the trigger is set to "hover" the timer is established on hover, but if the thing that you are hovering over is a button that you click to go to another page, then the timer is not cleared. The issue with this is most visible if the component is a keepalive one, so it is not destroyed. The timer fires after the page is gone, leaving a rogue popover at the top of the page. 

I therefore implemented a "beforeDestroy" function and "deactivated" function to handle both keepalive and non keepalive components. Both of these call the cleanup function which cancels the timer and makes sure the popover isn't displayed.

### DIFF
--- a/packages/popover/src/main.vue
+++ b/packages/popover/src/main.vue
@@ -123,11 +123,11 @@ export default {
     }
   },
 
-  beforeDestroy: function beforeDestroy() {
+  beforeDestroy() {
     this.cleanup();
   },
 
-  deactivated: function deactivated() {
+  deactivated() {
     this.cleanup();
   },
 
@@ -195,10 +195,9 @@ export default {
       this.$emit('after-leave');
       this.doDestroy();
     },
-    cleanup: function cleanup() {
+    cleanup() {
       if (this.openDelay) {
         clearTimeout(this._timer);
-        this.showPopper = false;
       }
     }
   },


goldengecko:
The failure of the test seems to be completely unrelated to the change I made since I only edited the documentation which was missing some format options which are supported in utils/date.js

ziyoung:
Which issue does this pr fix? Could you please provide a demo?

goldengecko:
The first issue is simply demonstrated:
```
<el-popover content="Something" trigger="hover" :open-delay="5000">
  <el-button>Click me</el-button>
</el-popover>
```
**Expected**: you need to hover over the button for 5000ms before the popover shows and there is no way to show the popover before the open delay is complete.
**Actual**: the popover shows immediately when you click the button.
**Fixed by**: making the popover only show on focus when the trigger is set to `focus` or `click`

The second issue is slightly more work. Set up your router to have a page kept alive:
```
<keep-alive include="mypage">
    <router-view>
    </router-view>
</keep-alive>
```
Now in the `mypage` component, have a popover on a button which takes you to a different page:
```
<el-popover content="Something" trigger="hover" :open-delay="5000">
  <el-button @click="$router.push('/anotherpage')">Click me</el-button>
</el-popover>
```
**Expected**: when you click the button, the popover will not show.
**Actual**: when you click the button, the popover is shown at the top left corner of the browser window (presumably since its anchor element is missing)
**Fixed by**: ensuring that when the component is deactivated or destroyed, any pending timer is destroyed and the popover is hidden. To be even more thorough, the this.showPopper = false; line could even be moved below the if statement to clean up in other cases like manual display.

ziyoung:
add a issue link: https://jsfiddle.net/dpc78qb4/

packages/popover/src/main.vue
```diff
@@ -186,6 +194,12 @@ export default {
     handleAfterLeave() {
       this.$emit('after-leave');
       this.doDestroy();
+    },
+    cleanup: function cleanup() {
+      if (this.openDelay) {
+        clearTimeout(this._timer);
+        this.showPopper = false;
```
ziyoung: Popover component uses `vue-popper` mixin which contains `beforeDestroy` and `deactivated` hooks. In this function, there's no need to reset `showPopper`.

https://github.com/ElemeFE/element/blob/0e7a7d345f508531cfb170a613a6ece1a2c398b7/src/utils/vue-popper.js#L188-L199

packages/popover/src/main.vue
```diff
@@ -123,6 +123,14 @@ export default {
     }
   },
 
+  beforeDestroy: function beforeDestroy() {
```
ziyoung: Use es6 coding style.

PR #13285, opened 2018-11-03
Select: add `filter-change` event

st-sloth:

### Description

In some cases if the application needs to use the search query entered in a filterable select, it needs to have access to this value. Emitting value of changed query solves that.

### DIFF
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -365,6 +365,12 @@
         }
       },
 
+      query(val, oldVal) {
+        if (this.filterable && val !== oldVal) {
+          this.$emit('filter-change', val);
+        }
+      },
+
       visible(val) {
         if (!val) {
           this.broadcast('ElSelectDropdown', 'destroyPopper');
--- a/examples/docs/en-US/select.md
+++ b/examples/docs/en-US/select.md
@@ -332,7 +332,7 @@ Multiple select uses tags to display selected options.
       :value="item.value">
     </el-option>
   </el-select>
-  
+
   <el-select
     v-model="value11"
     multiple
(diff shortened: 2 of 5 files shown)


st-sloth:
It is not about custom filtering logic. In fact it's about keeping the internal filtering logic but letting the application know that the filtering query changed, to allow some side functionality to get access to the query value. For instance, it's a select of complex objects, and there is a button to create a new object that opens a form. And it is desirable for the form field that corresponds to label in the select to already contain the search query that user entered.

It can be done
* accessing the input value through internal component structure
* using `filter-method` / `remote-method` and reimplementing internal filtering logic

but it would be much simpler just getting the search query via event.

wacky6:
@ziyoung What do you think?

packages/select/src/select.vue
```diff
@@ -365,6 +365,12 @@
         }
       },
 
+      query(val, oldVal) {
+        if (this.filterable && val !== oldVal) {
```
wacky6: val & oldVal comparison is unnecessary, vue already guarantees that when you use watch.

st-sloth: Indeed, thank you. Removed.

ziyoung:
We should not add this feature. 

Sometimes, the query of Select is not the search text user inputs.

1. Select one option called `Option3`. Now the query is `Option3`. After I click the input box again,  all options display.
kapture 2018-11-06 at 11 31 53

2. I type `Option3` in the input box and only one option displays.
kapture 2018-11-06 at 11 36 06

The `filter-change` event is not reliable. 😅

st-sloth:
> We should not add this feature.
The feature by itself or the current broken unreliable implementation?

After you had wrote that I've spent several hours trying to fix the problem you pointed out, but I can't really wrap my head around the data flow in the component so I haven't finished yet.
But unfortunately I don't have much spare time at the moment so I'll update PR some time later.

island205:
I'll close this PR since it's been a while since it's been opened. Feel free to reopen if you have updates on this

PR #13846, opened 2018-12-21
Select: add suffix slot to select component

vladislav-loza:
to create a specific actions like "Create", "Edit", "Remove" for elements in select box


island205:
it‘s not a common needs.

PR #13962, opened 2019-01-02
Form: custom messages for form validator

wesley6j:
Set custom error messages globally for Element UI forms:
```javascript
const messages =  {required: 'Please enter %s'}
Vue.use(Element, {validatorMessages: messages)
```

I18n compatible error messages can be achived with messages like this:

```javascript
import { t } from 'element-ui/src/locale';
const fieldLabel = field => t(`fields.${field}`) || field;

const message = (field, errorType, op = {}) => {
  let options = {...op, f: fieldLabel(field)};
  let ret = t(`errors.${field}.${errorType}`, options);
  if (ret) return ret;

  return t(`errors.${errorType}`, options);
};
export default {
  default(field) {
    return message(field, 'default');
  },
  required(field) {
    return message(field, 'required');
  }
}
```

### DIFF
--- a/packages/form/src/form-item.vue
+++ b/packages/form/src/form-item.vue
@@ -39,6 +39,7 @@
   import emitter from 'element-ui/src/mixins/emitter';
   import objectAssign from 'element-ui/src/utils/merge';
   import { noop, getPropByPath } from 'element-ui/src/utils/util';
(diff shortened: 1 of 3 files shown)


island205:
This is not a good solution to custom validator's message.  PR closed, if you have any update, please reopen this PR

PR #14133, opened 2019-01-17
Slider: fix max too large the browser crash bug

### LINKED ISSUE #14132: [Bug Report] in Slider, if the value of max is too large(like 10000000), the browser will crash
### Element UI version
2.4.11

### OS/Browsers version
osx/chrome 71

### Vue version
2.6.0-beta.1

### Reproduction Link
[https://jsfiddle.net/KarthusLorin/25uwnk6c/4/](https://jsfiddle.net/KarthusLorin/25uwnk6c/4/)

### Steps to reproduce
change the value of max to 10000000

### What is Expected?
the browser can run

in Slider, if the value of max is too large(like 10000000), the browser will crash

KarthusLorin:
if the value of max is too much like 10000000 the browser will crash because html has too much line `<!---->` (if the value of `v-if` is false , will leave `<!---->`, because of `v-for`, there are too much `<!---->`) , let max to outer will reduce the number of `<!---->` to 1

- closes #14132

### DIFF
--- a/packages/slider/src/main.vue
+++ b/packages/slider/src/main.vue
@@ -45,13 +45,14 @@
         ref="button2"
         v-if="range">
       </slider-button>
-      <div
-        class="el-slider__stop"
-        v-for="(item, key) in stops"
-        :key="key"
-        :style="vertical ? { 'bottom': item + '%' } : { 'left': item + '%' }"
-        v-if="showStops">
-      </div>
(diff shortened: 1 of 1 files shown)


iamkun:
Issues without a valid reproduction link will be closed directly.

你提供的链接不能复现这个问题

PR #14146, opened 2019-01-18
Table: make toggleAllSelection method an instance method(#14075)

### LINKED ISSUE #14075: [Bug Report] el-table method toggleAllSelection on mounted event works only for the last call
### Element UI version
2.4.11

### OS/Browsers version
OSX / Chrome Version 71.0.3578.98 (Official Build) (64-bit)

### Vue version
2.5.22

### Reproduction Link
[https://jsfiddle.net/9oxt38d6/](https://jsfiddle.net/9oxt38d6/)

### Steps to reproduce
With 2 tables, if I want to check all rows in both, I call the method `toggleAllSelection` on mounted, only the last call works

```js
mounted () {
   this.$refs.multipleTable1.toggleAllSelection() // dont work
   this.$refs.multipleTable2.toggleAllSelection() // works
}
```

### What is Expected?
All tables should be with all rows checked.
<details>
I want to start the tables with all rows selected, so, since I don't have a prop to set the wich rows should be selected, I'm trying to use the method `toggleAllSelection` on mounted phase of the component
</details>

The last call of `toggleAllSelection` is the only one that works.
<details>
Dont matter the data or any other config, only the last one works. If I put a `setTimeout` works.

```js
mounted () {
   this.$refs.multipleTable1.toggleAllSelection() // works
   setTimeout(() => {
      this.$refs.multipleTable2.toggleAllSelection() // works
   },  100);
}
```

</details>

LilyWakana:
fix #14075 .
All instances of TableStore share the same toggleAllSelection method in TableStore.prototype.mutations.Since the toggleAllSelection method is a debounce function which create a closure, all TableStore's instances share the same Lexical scope while they call the toggleAllSelection method. The time interval between two tables call `toggleAllSelection` is less than the debouncing delay(10), so the first call doesn't work.

This commit make the actual toggleAllSelection method an instance method.

### DIFF
--- a/packages/table/src/table-store.js
+++ b/packages/table/src/table-store.js
@@ -112,6 +112,35 @@ const TableStore = function(table, initialState = {}) {
     selectOnIndeterminate: false
   };
 
+  this._toggleAllSelection = debounce(10, function(states) {
+    const data = states.data || [];
+    if (data.length === 0) return;
+    const selection = this.states.selection;
+    // when only some rows are selected (but not all), select or deselect all of them
+    // depending on the value of selectOnIndeterminate
+    const value = states.selectOnIndeterminate
+      ? !states.isAllSelected
+      : !(states.isAllSelected || selection.length);
+    let selectionChanged = false;
+    data.forEach((item, index) => {
+      if (states.selectable) {
+        if (states.selectable.call(null, item, index) && toggleRowSelection(states, item, value)) {
+          selectionChanged = true;
+        }
+      } else {
+        if (toggleRowSelection(states, item, value)) {
+          selectionChanged = true;
+        }
+      }
+    });
+    const table = this.table;
+    if (selectionChanged) {
+      table.$emit('selection-change', selection ? selection.slice() : []);
+    }
+    table.$emit('select-all', selection);
+    states.isAllSelected = value;
+  });
+
   for (let prop in initialState) {
     if (initialState.hasOwnProperty(prop) && this.states.hasOwnProperty(prop)) {
       this.states[prop] = initialState[prop];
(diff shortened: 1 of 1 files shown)


packages/table/src/table-store.js
```diff
@@ -112,6 +112,35 @@ const TableStore = function(table, initialState = {}) {
     selectOnIndeterminate: false
   };
 
+  this._toggleAllSelection = debounce(10, function(states) {
```
island205: this.toggleAllSelection=debunce(10, this._toggleAllSelection)

packages/table/src/table-store.js
```diff
@@ -347,36 +376,9 @@ TableStore.prototype.mutations = {
...
-    });
-
-    const table = this.table;
-    if (selectionChanged) {
-      table.$emit('selection-change', selection ? selection.slice() : []);
-    }
-    table.$emit('select-all', selection);
-    states.isAllSelected = value;
-  })
+  toggleAllSelection(state) {
```
island205: _toggleAllSelection(state) {

island205:
@LilyWakana please move _toggleAllSelection to the prototype method

packages/table/src/table-store.js
LilyWakana: 执行toggleAllSelection方法时commit的操作是`toggleAllSelection`，所以TableStore.prototype.method要有一个同名方法toggleAllSelection。

packages/table/src/table-store.js
LilyWakana: 这里是为了避免每次实例化时都创建一个匿名函数吗？这里可以把传入debounce的函数声明成一个常量，这样在原型链也查找不到了。

island205:
Great work, Thank you for your contributions!

island205:
by the way:

```
let o = {
  x() {
    console.log('x')
  }
}
let x = function () {
  console.log('y')
  this.x()
}
x.call(o)
```

The best solution is: put `toggleAllSelection` to the `TableStore.prototype`,  add `this.toggleAllSelection=debunce(10, this.toggleAllSelection)` to the `TableStore` constructor funciton.

LilyWakana:
> by the way:
> 
> ```
> let o = {
>   x() {
>     console.log('x')
>   }
> }
> let x = function () {
>   console.log('y')
>   this.x()
> }
> x.call(o)
> ```
> The best solution is: put `toggleAllSelection` to the `TableStore.prototype`, add `this.toggleAllSelection=debunce(10, this.toggleAllSelection)` to the `TableStore` constructor funciton.

got it.thank you.

PR #14193, opened 2019-01-23
Autocomplete: Add custom key

jwm0:
Added custom key prop for `li` tag in autocomplete input. Index isn't a good key when data is dynamic and changes often. Vue would often lose track of the node and mix component state, resulting in unwanted behaviour. I propose a `customKey` prop that lets you specify the key yourself with a fallback to index when non is provided.


island205:
good practice, but we have no plan to support this at this time.

PR #14269, opened 2019-01-30
Autocomplete: prop to highlight first item

YamenSharaf:

This PR adds a prop to enable a user to have the first item highlighted by default when using remote search.

I have also applied this to the Algolia search component in the header for quick demoing, and because I feel this is a speedier approach for people navigating the docs. Could be removed though.

This addresses/ fixes this issue
https://github.com/ElemeFE/element/issues/14270

### DIFF
--- a/examples/components/search.vue
+++ b/examples/components/search.vue
@@ -36,81 +36,81 @@
 </template>
 
 <style>
-.algolia-search {
-  width: 450px !important;
+  .algolia-search {
+    width: 450px !important;
+  
+    &.is-empty {
+      .el-autocomplete-suggestion__list {
+        padding-bottom: 0;
+      }
+    }
 
-  &.is-empty {
     .el-autocomplete-suggestion__list {
-      padding-bottom: 0;
+      position: static !important;
+      padding-bottom: 28px;
     }
-  }
 
-  .el-autocomplete-suggestion__list {
-    position: static !important;
-    padding-bottom: 28px;
-  }
-
-  li {
-    border-bottom: solid 1px #ebebeb;
-
-    &:last-child {
-      border-bottom: none;
+    li {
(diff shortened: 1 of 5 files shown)


YamenSharaf:
> `highlightIndex` should be reset after suggestions are fetched in `getData`.
> 
> [element/packages/autocomplete/src/autocomplete.vue](https://github.com/ElemeFE/element/blob/8664fc5fc1b92f451e3d46380e348ad86712fc12/packages/autocomplete/src/autocomplete.vue#L182)
> 
> Line 182 in [8664fc5](/ElemeFE/element/commit/8664fc5fc1b92f451e3d46380e348ad86712fc12)
> 
>              this.suggestions = suggestions; 
> IMO, initialize `highlightIndex` to `0 / -1` is not necessary if `getData` resets it, so there is no need for `getHighlightedIndex` method.

Not sure what you mean that `getHighlightedIndex` is unnecessary. It's only there to toggle the behavior based on the props passed. Otherwise, how would it be toggleable to prevent breaking exisitng behavior?

I thought what you meant is that 
```
this.suggestions = suggestions;
```

should be this

```
this.suggestions = suggestions;
this.highlightedIndex = this.getHighlightedIndex();
```

abdallanayer:
this feature is also available on vuetifyjs https://vuetifyjs.com/en/components/autocompletes
auto-select-first

wacky6:
```
this.suggestions = suggestions;
this.highlightedIndex = this.getHighlightedIndex();
```

This bit is correct.

---

If you are going to have `getHighlightedIndex` in only one place, then it does not need to have a function. You can just write 

```
this.highlightedIndex = this.highlightFirstItem ? 0 : -1;
```

YamenSharaf:
@wacky6 You're right I had missed that, thanks. Fixed accordingly

PR #14278, opened 2019-01-31
MessageBox: Correct type definition

NateScarlet:
`MessageBoxData` may be a action string
https://github.com/ElemeFE/element/blob/b2f0958089a197907afdbea450a9faec69b48f02/packages/message-box/src/main.js#L57-L62

### DIFF
--- a/types/message-box.d.ts
+++ b/types/message-box.d.ts
@@ -2,7 +2,7 @@ import Vue, { VNode } from 'vue'
 import { MessageType } from './message'
 
 export type MessageBoxCloseAction = 'confirm' | 'cancel' | 'close'
-export type MessageBoxData = MessageBoxInputData
+export type MessageBoxData = MessageBoxInputData | MessageBoxCloseAction
 
 export interface MessageBoxInputData {
   value: string,


dancampbell691:
This is causing a TypeScript compile error for me:

> TS2339: Property 'value' does not exist on type 'MessageBoxData'.
>   Property 'value' does not exist on type '"confirm"'.

Pretty straight forward usage:
```javascript
let newFileName = await this.$prompt('File Name (Change if needed)', {
            closeOnClickModal: false,
            closeOnPressEscape: false,
            inputValue: file.name
        });
let fileName = encodeFileName(newFileName.value);
```
(The compile error points to the last line above.)

If I change the last line to simply `let fileName = encodeFileName(newFileName);` then it will compile, but of course I get a string of `[object] [object]`.  However, if I look at newFileName in the debugger in runtime, I can see that .value is definitely a property as expected.  In other words, the pure JS side works as expected, but the type change is preventing compile.

If I remove the change in this PR, it works as expected.  Am I doing something wrong, or is there truly a problem in the PR?

NateScarlet:
@dancampbell691 
You should add something like

```TypeScript
if (typeof newFileName == 'string'){
   throw new Error('Wrong result type')
}
```

See [Type Guards and Differentiating Types](https://www.typescriptlang.org/docs/handbook/advanced-types.html)

dancampbell691:
Thanks, that worked:
```javascript
        let newFileName = await this.$prompt('File Name (Change if needed)', {
            closeOnClickModal: false,
            closeOnPressEscape: false,
            inputValue: file.name
        });
        let fileName: string = '';
        if (typeof newFileName == 'string'){
            throw new Error('Wrong result type')
            }
        else {
            fileName = encodeFileName(newFileName.value);
        }
```

But that means that my one simple line `let fileName = encodeFileName(newFileName.value);` is now replaced by 7 lines.  Seems like this should be mentioned in the documentation or at least the changelog, since it's a breaking change.

Thanks so much for the quick and effective response.

NateScarlet:
Runtime code not have any change
It have a wrong type definition before that can save lines in your case, but cause error elsewhere

PR #14299, opened 2019-02-03
tree node-keys

Shinemax1:
e.g
```js
let data = {
  current: {
     id: 1,
     label: '栗子'
  },
  children: []
}
node-key: 'current.id'
```


ziyoung:

PR #14325, opened 2019-02-11
chore: import version from package.json

zy410419243:
- Synchronize version from `package.json` instead of updating manually.

- Changed in `build/webpack.conf.js` because of eslint problem.

P.S. Why not trying `git hooks` before commit?


ziyoung:
这个版本号是自动生成的，没有必要这样做。

PR #14660, opened 2019-03-10
Table: equal behaviour for height and max-height

### LINKED ISSUE #13751: The [Bug Report] Table attribute max-height uses a string without scrollbars
### Element UI version
2.4.11

### OS/Browsers version
win10 / chrome  70.0.3538.77

### Vue version
2.5.21

### Reproduction Link
[https://jsfiddle.net/Dingii/cp1u03ye/1/](https://jsfiddle.net/Dingii/cp1u03ye/1/)

### Steps to reproduce
Table属性max-height使用字符串设置，如150px，表格就不出现滚动条，如果是设置150则显示正常，
希望能和height设置一直，height属性设置150px和150表现是一样的

### What is Expected?
max-height使用带单位的字符串也能正常显示滚动条

使用带单位的值显示不正常

arthurdenner:
Closes #13751.

This PR is an attempt to close the referenced issue since I'm not sure if it breaks something...

Since the `max-height` is calculated [this way](https://github.com/ElemeFE/element/blob/dev/packages/table/src/table.vue#L510), if you specify it with a unit other than `px` (an integer), it doesn't calculate properly, therefore, doesn't show the scrollbar.

I looked into _why_ it's calculated like that and couldn't figured out, so I tried to replace it with the same calculation used for `height` and it worked as expected. Even [this example](http://element.eleme.io/#/en-US/component/table#fluid-height-table-with-fixed-header-and-columns), created in the [PR that originated the `max-height` feature](https://github.com/ElemeFE/element/pull/1674), just worked as before.

I checked the documentation after changing the calculation and all the examples are working as before, so it seems to me that the change doesn't break anything, but please review it carefully since I'm not aware of all the edge cases covered by the current calculation.

### DIFF
--- a/packages/table/src/table.vue
+++ b/packages/table/src/table.vue
@@ -507,9 +507,7 @@
           };
         } else if (this.maxHeight) {
           return {
-            'max-height': (this.showHeader
-              ? this.maxHeight - this.layout.headerHeight - this.layout.footerHeight
-              : this.maxHeight - this.layout.footerHeight) + 'px'
+            'max-height': this.layout.bodyHeight ? this.layout.bodyHeight + 'px' : ''
           };
         }
         return {};


ziyoung:
@arthurdenner It works for me.

ziyoung:
I checked the commit records of `Table.vue` and I think it is ok to do so. This is git blame of `Table.vue` https://github.com/ElemeFE/element/blame/595271f46332287cfcd42385c5b9ba05927e951d/packages/table/src/table.vue#L339

zhuxianbin:
There is another bug, set the table of max-height and the height is wrong.

askart:
Can confirm, all tables' height broken

cjsky666:
There is another bug, only set the table of max-height and the height is wrong [](url)

ziyoung:
When max-height is string, it will cause another bug. https://codepen.io/ziyoung/pen/qwvryL

fangwenzheng88:
There is another bug，will change it back in the next release

kobezhu:
why did you do that,this leads to bugs

Y-Dude:
set max-height  then
max-height was forcibly modified to 60px
image

gavin1990:
bodyHeight returns the default empty height in case of data loading delay.
image
image

XMoonLights:
> set max-height then
> max-height was forcibly modified to 60px
> image

I also have this problem, the max-heigth setting is successful, but the wrapper's max-heigth is fixed at 77px.

LoseyourseIf:
I also have this problem height:'80vh' does not work ! so I used old version '2.7.2'  it's ok

PR #15008, opened 2019-04-08
fix: Fix Notification title word break

### LINKED ISSUE #14893: [Bug Report] Notification spacing
### Element UI version
2.7.0

### OS/Browsers version
MacOS 10.13.6 / Chrome 73.0.3683.86

### Vue version
2.6.10

### Reproduction Link
[https://jsfiddle.net/r4qnuxjs/](https://jsfiddle.net/r4qnuxjs/)

### Steps to reproduce
Create a notification with a title long enough to fill the entire top row.

### What is Expected?
There should be an even amount of spacing between the notification content and the close button.

The title runs right into the close button, making it unclear there's an X to click sometimes.

Screen Shot 2019-03-28 at 13 52 32

iamkun:
close #14893

### DIFF
--- a/packages/theme-chalk/src/common/var.scss
+++ b/packages/theme-chalk/src/common/var.scss
@@ -370,7 +370,7 @@ $--message-danger-color: $--color-danger !default;
 /* Notification
 -------------------------- */
 $--notification-width: 330px !default;
-$--notification-padding: 14px 26px 14px 13px !default;
+$--notification-padding: 14px 28px 14px 13px !default;
 $--notification-radius: 8px !default;
 $--notification-shadow: $--box-shadow-light !default;
 $--notification-border-color: $--border-color-lighter !default;
--- a/packages/theme-chalk/src/common/var.scss
+++ b/packages/theme-chalk/src/common/var.scss
@@ -370,13 +370,14 @@ $--message-danger-color: $--color-danger !default;
 /* Notification
 -------------------------- */
 $--notification-width: 330px !default;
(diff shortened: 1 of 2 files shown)


breadadams:
@iamkun, I don't think that this is an appropriate fix. You can't just split words up for **everyone**'s notifications. It looks bad, and I _know_ my client will complain about this.

The correct solution would be to adjust the padding so that the text never reaches the X. I can sort this is if you want. Please re-open my original issue.

iamkun:
@breadadams agreed

breadadams:
Cool @iamkun, it's looking better! However given the right amount of characters the text could still run just against the side of the ✖️ icon:

image

I'd add ~8px of padding-right to `.el-notification__title`, leaving the main notifications padding as is.

iamkun:
I think this 2px fix would be enough for most cases.

breadadams:
Sure thing @iamkun, could be...

I couldn't actually get it to happen with anything in english other than random characters, however a client of mine managed to with a Spanish string. Let's see how the 2px fix works and if necessary I'll open a hotfix 😉

breadadams:
Very good point @wacky6, I'd never tried a notification without a title.

The right padding needs increasing a bit more. From a UI point of view I think it's better with a bigger gap like in the above screenshot, it separates the concerns of the "content" and the "action area" better (imo).

So maybe changing from `28px` → `38px`?

iamkun:
@breadadams @wacky6 Updated based on reviews. 😀

Issue #15010, opened 2019-04-08
[Bug Report] Table header's sort conditon gets lost when listening to sort-change events


joelxr:
### Element UI version
2.7.2

### OS/Browsers version
any

### Vue version
2.6.10

### Reproduction Link
[https://jsfiddle.net/joelxr/k8t26cd9/1/](https://jsfiddle.net/joelxr/k8t26cd9/1/)

### Steps to reproduce
1. Click on any sortable header once (FirstName column for example), to set column order to ascending.
2.  Click on the same header for the second time to set sorting condition to descending.
3. Click on the same header once more to unset the sorting condition.

### What is Expected?
On the method handleSortChanged that listens to "sort-change" event.
After the third interaction, I would like to have null value only in the "order" argument to keep a track on my sort object.

For me it makes more sense on click 3 the incoming values on the handler:

 Click 3: column: {...}, prop: 'name', order: null

Because, this way, I able to know that a sorting condition (order) on a specific column (prop)  has been set to null and then apply it on my remote request.

### What is actually happening?
 When the header is clicked the handler to "sort-change"  will alternate in the folowing values:

Click 1: column: {...}, prop: 'name', order: 'ascending'
 Click 2: column: {...}, prop: 'name', order: 'descending'
 Click 3: column: null, prop: null, order: null

And so it goes...

PR #15012, opened 2019-04-08
Table: Fix sort-change behaviour when sort condition is null (#15010)

joelxr:
As explained in #15010 this fixes the behavior of `sort-change` events on Tables.
Is expected to a handler of this particular event to track all the changes of the sorting condition to a column.

See the issue for more details.

### DIFF
--- a/packages/table/src/table-header.js
+++ b/packages/table/src/table-header.js
@@ -501,8 +501,6 @@ export default {
 
       if (!order) {
         sortOrder = column.order = null;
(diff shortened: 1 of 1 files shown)


zhaofy:
这个问题处理并将在2.9.0发布，那我想知道2.9.0什么时候更新呢

ziyoung:
> 这个问题处理并将在2.9.0发布，那我想知道2.9.0什么时候更新呢

月底。

PR #15027, opened 2019-04-09
Dialog: Fix dialog body correctly break long words

### LINKED ISSUE #14855: [Bug Report] words can not be break in the dialog
### Element UI version
2.6.3

### OS/Browsers version
macOS 10.13.4 / Chrome 72.0.3626.121

### Vue version
2.6.10

### Reproduction Link
[https://jsfiddle.net/jds4a9gk/7/](https://jsfiddle.net/jds4a9gk/7/)

### Steps to reproduce
click the word " Click Dialog "

### What is Expected?
words in the dialog can be break into different line

the position of some words is out of range

iamkun:
fix #14855

### DIFF
--- a/packages/theme-chalk/src/dialog.scss
+++ b/packages/theme-chalk/src/dialog.scss
@@ -67,6 +67,7 @@
     padding: 30px 20px;
     color: $--color-text-regular;
     font-size: $--dialog-font-size;
+    word-break: break-all;
   }
 
   @include e(footer) {


d--j:
This is too aggressive. `word-break: break-all;` breaks apart words even when it is not necessary. 
It is common to nest e.g. an `<el-table>` inside an `<el-dialog>`. Then all columns of that table inherit the `word-break: break-all;` and you explicitly need to overwrite that so that the layout does not gets broken.

See https://jsfiddle.net/5c9x3p6e/
Edit_fiddle_-_JSFiddle

Maybe  [`overflow-wrap: break-word;`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap)  is better – or leave #14855 to be fixed on a case by case basis (when you have long lines in a tiny `<el-dialog>`, add your own CSS like in https://jsfiddle.net/dLsnc8v4/ )

PR #15075, opened 2019-04-11
Input: add show-word-count attribute

luckyCao:

### DIFF
--- a/packages/input/src/input.vue
+++ b/packages/input/src/input.vue
@@ -49,9 +49,9 @@
       <!-- 后置内容 -->
       <span
         class="el-input__suffix"
-        v-if="$slots.suffix || suffixIcon || showClear || showPassword || showWordCountVisible || validateState && needStatusIcon ">
+        v-if="getSuffixVisible()">
         <span class="el-input__suffix-inner">
-          <template v-if="!showClear || !showPwdVisible || !showWordCountVisible">
+          <template v-if="!showClear || !showPwdVisible || !isWordLimitVisible">
             <slot name="suffix"></slot>
             <i class="el-input__icon"
               v-if="suffixIcon"
@@ -66,7 +66,7 @@
             class="el-input__icon el-icon-view el-input__clear"
             @click="handlePasswordVisible"
           ></i>
-          <span v-if="showWordCountVisible" class="el-input__count">
+          <span v-if="isWordLimitVisible" class="el-input__count">
             <span class="el-input__count-inner">
               {{ textLength }}/{{ upperLimit }}
             </span>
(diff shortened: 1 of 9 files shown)


wacky6:
Prefer not to have word count in mainline.

The definition of a word may differ. In English one word consists of multiple characters. In Chinese, a word usually means a single glphy.

Also, JavaScript `.length` is the number of UTF-16 code units, it does not work for complex characters (surrogate pairs). (e.g. `'😀'.length === 2`, `'𥑮'.length === 2`). To get the number of code points (real character / glyph), use `[...string].length`.

---

`maxlength` is passed to native input and already limits `value.length`, so `isExceed` will never be true. Also, `maxlength` should be coerced into a number in `upperLimit`.

Shouldn't word count work without `maxlength` ? User may only want to hint character count rather than to enforce a strict limit. 

Does it conflict with `clearable` / other suffix slot usage?

luckyCao:
Thanks remind complex characters, will fix it.
When initial value's length is larger then upperLimit, isExceed is true. 
It's not conflict with clearable / other suffix slot

packages/input/src/input.vue
```diff
@@ -48,9 +49,9 @@
       <!-- 后置内容 -->
       <span
         class="el-input__suffix"
-        v-if="$slots.suffix || suffixIcon || showClear || showPassword || validateState && needStatusIcon">
+        v-if="$slots.suffix || suffixIcon || showClear || showPassword || showWordCountVisible || validateState && needStatusIcon ">
```
ziyoung: 优化一下。这块判断太复杂。

types/input.d.ts
```diff
@@ -81,6 +81,15 @@ export declare class ElInput extends ElementUIComponent {
...
   validateEvent: boolean
 
+  /** Whether the input is clearable */
+  clearable?: boolean
+
+  /** Whether to show password */
+  showPassword?: boolean
+
+  /** Whether to show wordCount when setting maxLength */
+  showWordCount?: boolean
```
SimonaliaChen: prefer showWordLimit

types/input.d.ts
```diff
@@ -81,6 +81,15 @@ export declare class ElInput extends ElementUIComponent {
   /** Whether to trigger form validatio */
   validateEvent: boolean
 
+  /** Whether the input is clearable */
+  clearable?: boolean
```
SimonaliaChen: is there any need to add question mark?

packages/input/src/input.vue
```diff
@@ -49,9 +49,9 @@
       <!-- 后置内容 -->
       <span
         class="el-input__suffix"
-        v-if="$slots.suffix || suffixIcon || showClear || showPassword || showWordCountVisible || validateState && needStatusIcon ">
+        v-if="$slots.suffix || suffixIcon || showClear || showPassword || isWordLimitVisible || (validateState && needStatusIcon) ">
```
SimonaliaChen: 个人认为这么复杂的判断语句 直接写在模板里是不是不太好维护？考虑放在计算属性里吧

ziyoung: 有 bug 的。slot 不是响应式的。

lvjiaxuan:
在设置了size的el-form里的el-input，使用show-word-count 有样式问题？？？？？？？

PR #15439, opened 2019-05-07
Table: fix sort icon

bezany:
Reproduce link: https://codepen.io/anon/pen/BeygzE
After click `Sort by Name` sort icon from `Date` column not hidden.
I found that need clear `sortingColumn.order` when change sort column.

### DIFF
--- a/test/unit/specs/table.spec.js
+++ b/test/unit/specs/table.spec.js
@@ -1753,6 +1753,7 @@ describe('Table', () => {
           vm.$nextTick(() => {
             expect(toArray(lastCells).map(node => node.textContent))
               .to.eql(['-100', '-95', '-92', '-92', '-80']);
+            destroyVM(vm);
             done();
           });
         });
--- a/test/unit/specs/table.spec.js
+++ b/test/unit/specs/table.spec.js
@@ -1758,6 +1758,41 @@ describe('Table', () => {
         });
       }, DELAY);
     });
+
+    it('sort correct change icon', async() => {
+      function assertSortIconCount($el, msg, count = 1) {
+        const sortIconCount = $el.querySelectorAll('th.ascending, th.descending').length;
+        expect(sortIconCount).to.equal(count, msg);
+      }
+
+      const vm = createVue({
+        template: `
+          <el-table ref="table" :data="testData" >
+            <el-table-column prop="name" sortable />
+            <el-table-column prop="release" sortable />
+            <el-table-column prop="director" sortable />
+            <el-table-column prop="runtime" sortable />
+          </el-table>
+        `,
+        data() {
+          return { testData: getTestData() };
+        }
+      });
+      await waitImmediate();
+      assertSortIconCount(vm.$el, 'sorting icon is not empty after mount', 0);
+      // manual click first column header
+      const elm = vm.$el.querySelector('.caret-wrapper');
+      elm.click();
+      await waitImmediate();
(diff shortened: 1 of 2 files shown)


ziyoung:
Please resolve conflict.

bezany:
@ziyoung Done! May be more elegant added fix [here](https://github.com/ElemeFE/element/blob/f6bab114e3f8126a942938cf59d1afe5b09d9337/packages/table/src/store/watcher.js#L276), but I don't fully understand code.

ziyoung:
> @ziyoung Done! May be more elegant added fix [here](https://github.com/ElemeFE/element/blob/f6bab114e3f8126a942938cf59d1afe5b09d9337/packages/table/src/store/watcher.js#L276), but I don't fully understand code.

What is your question?

packages/table/src/store/index.js
```diff
@@ -70,6 +70,9 @@ Watcher.prototype.mutations = {
   sort(states, options) {
     const { prop, order } = options;
     if (prop) {
+      if (states.sortingColumn) {
```
ziyoung: You can add these lines in `$nextTick` callback.

test/unit/specs/table.spec.js
```diff
@@ -1758,6 +1758,52 @@ describe('Table', () => {
         });
       }, DELAY);
     });
+
+    it('sort correct change icon', done => {
```
ziyoung: For asynchronous testing, use the async function.

test/unit/specs/table.spec.js
```diff
@@ -1758,6 +1758,52 @@ describe('Table', () => {
...
+        template: `
+          <el-table ref="table" :data="testData" >
+            <el-table-column prop="name" sortable />
+            <el-table-column prop="release" sortable />
+            <el-table-column prop="director" sortable />
+            <el-table-column prop="runtime" sortable />
+          </el-table>
+        `,
+
+        created() {
```
ziyoung: Why do you set `testData` in the created hook? You can set `testData` in data function.
```js
data() {
  return { testData: getTestData() }
}
```

island205:
@bezany Please rebase `dev` branch and resolve conflict.

test/unit/specs/table.spec.js
bezany: You right, thanks.

test/unit/specs/table.spec.js
bezany: I found utils function `waitImmediate`. Good!

packages/table/src/store/index.js
bezany: I moved fix in another function.

bezany:
@ziyoung Thank you for review! I improved code.
@island205 Did rebase and fix conflict.
Now fix in [`updateSort`](https://github.com/bezany/element/blob/fix/table_sort/packages/table/src/store/watcher.js#L269) function.

When I fix issue I saw, that code in [`handleSortClick` function](https://github.com/ElemeFE/element/blob/f6bab114e3f8126a942938cf59d1afe5b09d9337/packages/table/src/table-header.js#L473) have same code as in [`updateSort` function](https://github.com/ElemeFE/element/blob/f6bab114e3f8126a942938cf59d1afe5b09d9337/packages/table/src/store/watcher.js#L276). And property `sortingColumn.order` look as duplicate of `sortOrder`. May be planning this changes for future refactoring.

PR #15509, opened 2019-05-11
[Tree]:Fix bug about method setCurrentKey

tangzhenkang:
setCurrentKey(null) can't cancel the highlighting tree-node


WaSheep:
please resolve the conflict

PR #13198, opened 2018-10-28
Progress: add text status

ali-master:
Please make sure these boxes are checked before submitting your PR, thank you!

* [x] Make sure you follow Element's contributing guide ([中文](https://github.com/ElemeFE/element/blob/master/.github/CONTRIBUTING.zh-CN.md) | [English](https://github.com/ElemeFE/element/blob/master/.github/CONTRIBUTING.en-US.md) | [Español](https://github.com/ElemeFE/element/blob/master/.github/CONTRIBUTING.es.md)).
* [x] Make sure you are merging your commits to `dev` branch.
* [x] Add some descriptions and refer relative issues for you PR.

Hi there

With the current version of ElementUI, we can not use the special text inner circle progress content, and I have added this feature, for instance:

```javascript
<Progress :percentage="100" :stroke-width="10" :width="100" status="text" type="circle">Done</Progress>
```

And it will be:
screen shot 2018-10-29 at 1 53 09 am

I hope it will be useful.

PR #15601, opened 2019-05-17
Progress: revert adding text status

island205:
Reverts ElemeFE/element#13198

Because `status` and `text` has no relation, we will find another approach to show text.

### DIFF
--- a/examples/docs/en-US/progress.md
+++ b/examples/docs/en-US/progress.md
@@ -37,8 +37,7 @@ In this case the percentage takes no additional space.
 <el-progress type="circle" :percentage="80" color="#8e71c7"></el-progress>
 <el-progress type="circle" :percentage="100" status="success"></el-progress>
 <el-progress type="circle" :percentage="50" status="exception"></el-progress>
-<el-progress type="circle" :percentage="100" status="text">Done</el-progress>
-```
+``` 
 :::
 
 ### Attributes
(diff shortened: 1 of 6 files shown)


ffxsam:
@island205 Why was this reverted? With no console warnings about deprecation. This is a breaking change and broke our app when we upgraded from 2.6 to 2.10.

ffxsam:
@island205 Disappointing this was reverted with no console warnings about deprecation in future versions. This is a breaking change and broke our app when we upgraded from 2.6 to 2.10.

ffxsam:
This is a good read: https://semver.org/

>Given a version number MAJOR.MINOR.PATCH, increment the:

>MAJOR version when you make incompatible API changes,
>MINOR version when you add functionality in a backwards-compatible manner, and
>PATCH version when you make backwards-compatible bug fixes.
>Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

willvincent:
Totally, a breaking change should never happen without a major version number bump... If people are going to use this in production (and people do), they need to be able to rely on things not just randomly breaking in a minor or patch release... There should be no reason to have to pin a specific version number to ensure things don't randomly break.

island205:
@ffxsam @willvincent Very good advice, we will deal with breaking change seriously next time.

PR #15669, opened 2019-05-22
InputNumber: add type and readonly props (native input attrs)

leomp12:
New `type` property to edit native input type attribute, so the user may be able to change the default type (text) to tel (and maybe number, but it would change the default design).
Imho it's important to give better experience on mobile, showing the number keyboard on focus instead of default keyboard.

### DIFF
--- a/packages/input-number/src/input-number.vue
+++ b/packages/input-number/src/input-number.vue
@@ -35,6 +35,7 @@
       :max="max"
       :min="min"
       :type="type"
+      :readonly="readonly"
       :name="name"
       :label="label"
       @keydown.up.native.prevent="increase"
@@ -87,6 +88,7 @@
       },
       value: {},
       disabled: Boolean,
+      readonly: Boolean,
       size: String,
       controls: {
         type: Boolean,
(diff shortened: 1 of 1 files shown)


leomp12:
Edit: also adding `readonly` property.

wacky6:
emm. Shouldn't type="*" be used on el-input?

It does not make sense to use it with el-input-number which has increase/decrease control buttons.

leomp12:
Using el-input we lose the controls, using el-input-number we lose the numeric keyboard because of the input type...
Imo would be good to be able to handle both, so we still having the controls without making the mobile experience worse.

Even having the controls, sometimes the input will be focused (of course), and then it's fairly better to open a numeric keyboard specially when you're using a small mobile device.
Does not mean that I would use `type=number`, it's probably a better idea to use `type=tel`, which would open numeric keyboard on mobile, without adding native browser increase and decrease buttons :)

WaSheep:
Sorry, we have no plan to support this

WaSheep:
Sorry, we have no plan to support this

flysubmarina:
> Sorry, we have no plan to support this

So how can i make `<ElInputNumber> `readonly. Lets say we have a case when we want to **do not allow the user to enter data from the keyboard**, how can i achieve it? i have this problem now, looking forward for answer :)

PR #15761, opened 2019-05-28
Autocomplete: fix problem when input with Chinese input method and enabled select-when-unmatched

qazbnm456:
This PR fixes the problem when the Chinese input method is used with select-when-unmatched enabled. **Reproducible demo page: https://codepen.io/anon/pen/mYjKVP?editors=1111**

Expected behavior:
When input with the Chinese input method, the value won't be logged unless the user hits the Enter button after the compositionend event.

Actual behavior:
`handleKeyEnter` method doesn't check if we're still on composition and immediately logs the previous value before the compositionend event. If we want to get the actual value, we need to hit the Enter button one more time.

image


WaSheep:
Sorry, we have no plan to support this

qazbnm456:
@WaSheep It's not a PR for features. It's for solving an existing problem.

PR #16225, opened 2019-06-25
Upload: abort upload when `before-upload` return `Promise<false>`

### LINKED ISSUE #15939: [Feature Request] change `before-upload` behavior
### Existing Component
是

### Component Name
upload

### Description
I was using a function to prevent upload，code like this

```js
onBeforeUpload() {
  doSomthingBeforeUpload()
  return false
}
```

then one-day `doSomthingBeforeUpload` became a promise

```js
async onBeforeUpload() {
  await doSomthingBeforeUpload()
  return false
}
```

this won't stop the upload, because this function actually resolves with `false`, not rejected, so I have to throw an error in `doSomthingBeforeUpload` or onBeforeUpload.

I read the code https://github.com/ElemeFE/element/blob/1afce457e7a2ab3d5be83d0fc1973722ff6ba87f/packages/upload/src/upload.vue#L93-L111 and know that we thread `Blob` as a new upload file, and other values will upload the original file

Here is my propose: treat Promise&lt;false&gt; as a signal to prevent upload, I think it's better and will not break many use-cases.

If it's ok, I can make a pr to make this change

fisker:

this PR changes `Upload` behavior `before-upload`, when `before-upload` return `Promise&lt;false&gt` will abort upload

fixes #15939

**Notice: This PR has Breaking changes**

TODO:

docs except `zh-CN` and `en-US` are not updated

### DIFF
--- a/examples/docs/en-US/upload.md
+++ b/examples/docs/en-US/upload.md
@@ -358,7 +358,7 @@ on-success | hook function when uploaded successfully | function(response, file,
 on-error | hook function when some errors occurs | function(err, file, fileList) | — | —
 on-progress | hook function when some progress occurs | function(event, file, fileList) | — | — |
 on-change | hook function when select file or upload file success or upload file fail | function(file, fileList) | — | — |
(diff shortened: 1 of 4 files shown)


island205:

PR #16510, opened 2019-07-11
Table: fix clearFilter not work in some situation

liuqiang1357:
'clearFilter' not work if  a column init with a filtered-value prop and the filter in header has not been clicked. 

reproduction link: https://codepen.io/liuqiang1357/pen/xoMBWZ


liuqiang1357:
still not fixed.

PR #16643, opened 2019-07-19
Header: Remove default value for height property

manuasdf:
Remove default value for the header component property. Since the default value keeps the inline CSS active, it requires an !important to be overwritten. 

This change keeps the functionality of the property untouched. It allows header with a variable height.


luckyCao:

PR #16813, opened 2019-07-31
Dialog: fix body not render when invisible.

a631807682:

 Reproduction Link
---
https://codepen.io/a631807682/pen/zgwwvL?editors=1011

Steps to reproduce
---
1. open dialog
2. see the console

What is Expected?
---
footer ref is exist but body ref is not exist when dialog invisible.

What is actually happening?
---
footer and body ref is exist.

### DIFF
--- a/packages/dialog/src/component.vue
+++ b/packages/dialog/src/component.vue
@@ -25,7 +25,7 @@
             <i class="el-dialog__close el-icon el-icon-close"></i>
           </button>
         </div>
-        <div class="el-dialog__body" ><slot></slot></div>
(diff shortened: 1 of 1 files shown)


luckyCao:
dose this cause some problem?

a631807682:
@luckyCao i have some permission button in dialog using directive, i want to know if exist and then execute my logic, but it not render at first, so i should set visible to `true` to render it and check in `$nextTick`, and then i should set visible `false` , so my dialog should show and hidden.

a631807682:
1. i think body and footer render behavior should be consistent, both render or not.
2. if both not render, should use `prop` instead?

Issue #16865, opened 2019-08-01
[Bug Report] Time picker selected text is off by 1 when 12-hour format is used


mattheyan:
### Element UI version
2.11.1

### OS/Browsers version
Windows 10 / Google Chrome 75

### Vue version
2.6.10

### Reproduction Link
https://codepen.io/mattheyan/pen/gVxMbM

### Steps to reproduce
1. Click on the hours part in the text input.
2. If necessary, use the left or right keys until the hours part is selected.

### What is Expected?
Since the time format is a 12-hour format, the hour may only be 1 digit long. In this case, only that 1 digit should be selected.

### What is actually happening?
If the hour is less than 10, the hour portion is highlighted, but the colon separator is also highlighted. Also, when navigating to the minutes and seconds components, the selection is off by 1.

mattheyan:
I am able to implement a fix and will submit a PR for your review

Issue #16524, opened 2019-07-11
[Feature Request] Ability to toggle AM/PM in the time picker


mattheyan:
### Existing Component
Yes

### Component Name
TimePicker

### Description
We're working on incorporating element UI into our existing product. The product's current behavior, which we would like to keep, allows users to toggle a time field between AM and PM. The element time picker doesn't currently support this. Is this something you would consider adding? We would also be able to implement and submit a PR if the team's time is limited and you would be open to accepting a PR.

Thanks,
Bryan

iamkun:
pr welcomed

PR #16959, opened 2019-08-08
TimePicker: Add support for toggling AM/PM

### LINKED ISSUE #16866: [Bug Report] Time picker doesn't visually select the part being edited via directional keys
### Element UI version
2.11.1

### OS/Browsers version
Windows 10 / Google Chrome 75

### Vue version
2.6.10

### Reproduction Link
https://codepen.io/mattheyan/pen/gVxMbM

### Steps to reproduce
1. Click into the text input so that one of the time parts (hours, minutes, seconds) is selected
2. Press the up or down keys to increase or decrease the value

### What is Expected?
The value show increase or decrease, and that part of the time should remain selected

The value is increased or decreased, but the text selection disappears and the cursor moves to the end of the input.

mattheyan:
Feature request #16524 [Feature Request] Ability to toggle AM/PM in the time picker

Basic support for scrolling between AM and PM is implemented in this commit. However, in the course of implementing this there were a few other considerations that came up:

- When scrolling between AM and PM is enabled, should only 12 hours be shown in the hour selection list? If not, the controls are a bit redundant. If so, then this opens a few other questions:
  - What should happen when you scroll down at the end of the list, and wrap around to the top? As it currently stands, scrolling down from 11 AM moves to 12 PM. This makes sense logically, but it means that the hours and AM/PM scrollbars are "tethered" together, which is not consistent with existing behavior. The minute scrollbar doesn't currently cause the hour to increment or decrement when it wraps around (same with seconds and minutes). Another option that comes to mind is to have a prop to control whether this behavior should be in effect or not. If not, which would probably be the default, none of the controls would impact each other. If the prop was enabled, then they would all impact each other consistently. This would also mean that incrementing the minute from 59 to 0 would increment the hour, and incrementing the second from 59 to 0 would increment the minute. Its worth nothing that none of this should impact the default behavior of the time picker.
  - If we're toggling AM / PM and only showing 12 hours (and the wrap around behavior doesn't happen), then should the hours be listed as 1-12 instead of 12, 1, 2, .. 11? It seems that the logical ordering of these values is 12, 1, 2, ... 11, as it is currently, but when the hour numbers operate independently and don't wrap around, it may seem like the user is really just editing a number, and the current ordering might seem strange.
- It seems that scrolling between AM and PM complicates the use of selectable time ranges. The control shows you what hours, minutes, and seconds can be selected. However, when the hours are split in two and AM / PM is toggled separately, then you can no longer see what hours can be selected. It seems if you're going to restrict the times that can be selected, then using the time spinner with AM / PM toggling enabled is not a good choice of user interface, so it may be acceptable to do nothing about it.
- Currently when you enable AM/PM toggling, and arrow controls are NOT used, then all 24 hours are shown. It may be that these two options are co-dependent. You can't toggle AM/PM unless you use arrow controls, since scrolling is really better suited for the 24 hour format. Thoughts?

Changes to support AM/PM:
- Add 'toggleAmPm' prop
- Add 'amPm' to list of time parts to scroll
- Handle the conditionally available 'AM/PM' in selection range
- Determine whether the hour is zero padded based on the format
- Take into account the hour length in selection range logic
- Add 'has-am-pm' class to the time picker content element
- Only show 12 hours when toggling AM/PM with arrow controls
- Set the selection range after scrolling up or down

Also...

When using a 12-hour format, the selected characters in the time spinner are off by 1 when the hour is less than 10. This fixes the problem by determining the length that the selection should be for the hour part, based on the AM/PM mode and the hour value, and adjusts every part based on that value. This fixes issue #16865  - [Bug Report] Time picker selected text is off by 1 when 12-hour format is used

After scrolling up or down in the time picker's text box, the selected text is lost and the cursor goes to the end of the input (though scrolling continues to work). This fixes the problem by calling emitSelectRange at the end of the scrollDown method - `this.$nextTick(() => this.emitSelectRange(this.currentScrollbar));`. Fixes #16866 - [Bug Report] Time picker doesn't visually select the part being edited via directional keys

### DIFF
--- a/examples/docs/en-US/time-picker.md
+++ b/examples/docs/en-US/time-picker.md
@@ -169,6 +169,7 @@ Can pick an arbitrary time range.
 | name | same as `name` in native input | string | — | — |
 | prefix-icon | Custom prefix icon class | string | — | el-icon-time |
 | clear-icon | Custom clear icon class | string | — | el-icon-circle-close |
+| toggle-am-pm | whether an AM/PM selector should be rendered, only works with `<el-time-picker>` when the format includes AM/PM, and is best used in conjunction with `arrow-control` | boolean | — | false |
 
 ### Time Select Options
 | Attribute      | Description          | Type      | Accepted Values       | Default  |
--- a/packages/date-picker/src/basic/time-spinner.vue
+++ b/packages/date-picker/src/basic/time-spinner.vue
@@ -15,7 +15,7 @@
           v-for="(disabled, hour) in hoursList"
           class="el-time-spinner__item"
           :key="hour"
-          :class="{ 'active': hour === hours, 'disabled': disabled }">{{ ('0' + (amPmMode ? (hour % 12 || 12) : hour )).slice(-2) }}{{ getAmPm(hour, { prefix: ' ' }) }}</li>
+          :class="{ 'active': hour === hours, 'disabled': disabled }">{{ ('0' + (twelveHourClock ? (hour % 12 || 12) : hour )).slice(-2) }}{{ getAmPm(hour, { prefix: ' ' }) }}</li>
       </el-scrollbar>
       <el-scrollbar
         @mouseenter.native="emitSelectRange('minutes')"
@@ -79,7 +79,7 @@
             class="el-time-spinner__item"
             :class="{ 'active': hour === hours, 'disabled': hoursList[hour] }"
             v-for="(hour, key) in arrowHourList"
-            :key="key">{{ hour === undefined ? '' : ('0' + (amPmMode ? (hour % 12 || 12) : hour )).slice(-2) + (showAmPm ? '' : getAmPm(hour, { prefix: ' ' })) }}</li>
+            :key="key">{{ hour === undefined ? '' : ('0' + (twelveHourClock ? (hour % 12 || 12) : hour )).slice(-2) + (showAmPm ? '' : getAmPm(hour, { prefix: ' ' })) }}</li>
         </ul>
       </div>
       <div
@@ -161,6 +161,7 @@
         type: String,
         default: '' // 'a': am/pm; 'A': AM/PM
       },
+      twelveHourClock: Boolean,
       zeroPadHour: Boolean
     },
 
@@ -268,7 +269,7 @@
       },
 
       emitSelectRange(type) {
-        const hoursLen = !this.zeroPadHour && (this.date.getHours() % 12 || 12) < 10 ? 1 : 2;
+        const hoursLen = !this.zeroPadHour && (this.twelveHourClock ? (this.date.getHours() % 12 || 12) : this.date.getHours()) < 10 ? 1 : 2;
         const secondsOffset = this.showSeconds ? 3 : 0;
         if (type === 'hours') {
           this.$emit('select-range', 0, hoursLen);
--- a/packages/date-picker/src/basic/time-spinner.vue
+++ b/packages/date-picker/src/basic/time-spinner.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="el-time-spinner" :class="{ 'has-seconds': showSeconds }">
+  <div class="el-time-spinner" :class="{ 'has-seconds': showSeconds, 'has-am-pm': showAmPm }">
     <template v-if="!arrowControl">
       <el-scrollbar
         @mouseenter.native="emitSelectRange('hours')"
@@ -15,7 +15,7 @@
           v-for="(disabled, hour) in hoursList"
           class="el-time-spinner__item"
           :key="hour"
-          :class="{ 'active': hour === hours, 'disabled': disabled }">{{ ('0' + (amPmMode ? (hour % 12 || 12) : hour )).slice(-2) }}{{ amPm(hour) }}</li>
+          :class="{ 'active': hour === hours, 'disabled': disabled }">{{ ('0' + (amPmMode ? (hour % 12 || 12) : hour )).slice(-2) }}{{ getAmPm(hour, { prefix: ' ' }) }}</li>
       </el-scrollbar>
       <el-scrollbar
         @mouseenter.native="emitSelectRange('minutes')"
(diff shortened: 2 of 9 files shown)


mattheyan:
I just noticed that the test build failed due to test "selection range with 1 digit hour", which I added. It passes locally, so I would appreciate any pointers on how I could troubleshoot why it is failing in the build.

I'm awaiting feedback on these changes. Please let me know if there is anything needed from me, or if there is something I can do to help the process along. I suppose a demo in codepen would be useful, but I'm not sure how to make it possible to demonstrate the use of these changes that are not yet integrated into element, other than publishing my own build of the package somewhere. Using the local `index.vue`, the functionality can be previewed like so: `<el-time-picker v-model="time" format="h:mm A" arrow-control toggle-am-pm />`.

Also, of the comments above, I think the only thing that might require action is the question of how to handle wrap-around behavior. In other words, clicking down from 11 am currently advances to 12 pm. When controlling hours and AM/PM separately, this is inconsistent with how minutes and seconds work. However, I could introduce a prop to enable this wrap around behavior for minutes and seconds as well in a separate change, which I would like to do anyway, if this is acceptable.

Finally, there are a few other enhancements I'd like to make in order to be able to support the functionality our product needs. I intend make all of these changes in a backwards compatible way so that the default element behavior doesn't change unless you opt into these new options. I can create separate PRs for these changes, but if it would be more convenient to bundle them together then I can do that as well and include them in this PR or a new one. Whatever is best for you.
1. Option to increase the "higher" time part when incrementing from 59 to 0. For example, going from 59 minutes to 0 minutes would increment the hour. I mentioned this above with regards to AM/PM.
2. Option to specify an increment for increasing and decreasing a time value. For example, increment by 15 minutes at a time. In this scenario, if the value is not currently an increment of 15 it would move to the nearest 15 minute increment.
3. Option to control the number of rows shown when using arrow controls. The default would still be 3, but it could be changed to any value greater than or equal to 1.
4. Option to reverse the scrolling direction, such that scrolling down would decrease the value and up would increase the value. Using this option would make the scrolling direction more logical when using keyboard keys to scroll the value.

wacky6:
Some thoughts:

* Go from 11AM to 12PM is reasonable. AM/PM + 12 hour spinner combined is *equivalent* to hour selector. They should be internally represented as a single 24 hour attribute (0-23 hours), then determine each spinner's state (similar to computed property).
  * This is not wrap-around IMO
* Use typical 12 hour clock representation: https://en.wikipedia.org/wiki/12-hour_clock , 12am -> 11am -> 12pm -> 11pm -> 12am
* Don't worry about selectable range. AM/PM can be toggled as user wishes. The picker will refuse to close / reset if selected time is not allowed in `selectableRange`. Showing 12 disabled hour is acceptable
* AM/PM toggle should work without arrow-controls ? Have you encountered problems in implementing scrolling?
* off-by-one character is a nice fix. We would like to have it in a saperate PR if this one is not merged. 
* for new test cases, use `async await`, **do not use $nextTick**
* for "selection range with 1 digit hour", set a initial time. Otherwise picker will show the current time at execution, which might fail the test if it happens to be 9PM and KEYDOWN to 10PM.

---

I prefer call "Increase hour when minute goes from 59 to 0" a **carry** (analogue to math addition, where 35+47 carries the tens digit of 5+7, to 3+4+1).

The carry behavior seems uncessary. You can play with iOS or Android time spinners. I don't recall they carry minutes to hours. Better keep UI interaction similar to native ones (which users are already used to).

Control the number of rows sounds fine. But it is a low priority task IMO.

Can you elaborate on "reverse scrolling direction". Is it similar to macOS's natural scrolling direction? If this is the case, shouldn't this behavior be left for the operating system?

---

@iamkun  @island205 Are there any thoughts from UED on how to implement a AM/PM switch? I havn't seen AM/PM format for around a decade. :)

mattheyan:
@wacky6 I agree with your points above and will leave the current behavior of advancing form 11am -> 12pm. In terms of scrolling mode, I should have mentioned that it looked to me like it would be more complicated to filter down the list of hours to just 12 hours, since that work is done by the `getRangeHours` helper. I wonder if this AM / PM option could be targeted to arrow controls to keep it simple, since it seems "scrolling" AM / PM separately doesn't make much sense anyway. Adding support for filtering the hours list to AM or PM would complicate the code for a use case that may not be needed. Currently with this PR you can enable AM / PM toggling when you're using the default scrolling UI, but it also shows all 24 hours. Its a strange setup but it works. Is it OK if I leave it as-is?

The off by 1 fix is also in a separate PR #16867.

Hopefully your feedback about async/await and setting an initial time will help me fix the failing test. I will try to do that soon.

---

Also, thanks for the feedback on those additional changes I mentioned. I agree they are lower priority items, and would be opt-in to change from the default behavior. I mentioned them because we would need them to support our current product behavior. I like the *carry* terminology, that's perfect! The reverse scrolling option would allow you to reverse the scrolling direction so that it makes more sense for keyboard input, i.e. up increases the value and down decreases the value. This would only make sense if you're not showing a time spinner that looks like a clock-type interface, so you're just optimizing for keyboard input, so its a bit of an odd scenario. Also, definitely on the lower end in terms of priority.

wacky6:
Actually, you can introduce additional code to handle `getRangeHours`. Maybe a new computed hourList (that includes 24 hours representation, and 12 hour representation + AM/PM for display purpose).  As I have mentioned before, 12 hour + AM/PM should depend on the real 24 hour clock.

For now, you can **leave it as-is**. **But** add a note/warning stating that AM/PM toggle only works with arrow-control. We will see if AM/PM + scrolling is necessary based on feedback.

---

Nice work

mattheyan:
@wacky6 The tests are passing now, thanks for pointing out the problem! Also, the tests I added weren't using `nextTick` but I did change them from callbacks and `setTimeout` to instead using async/await and the `wait` helper, which is much better. Thanks for pointing that out too. I didn't rewrite all of the tests in the file but if you'd like me to do that I don't mind.

For the note/warning, were you thinking a comment in the source code or a `console.warn` message?

wacky6:
note/warning in BOTH documentation (example/docs/) and source code.

don't worry about non-english translation, we will do them later.

mattheyan:
I found a problem with how the selection range was being set for formats that don't zero-pad the hour, ex: "H:mm". I was incorrectly treating it as if it was a 12-hour format, so if the value was "13:00", it would set a selection of 1 character when it should have been 2. So, I added a new `twelveHourClock` field derived from the format. I also discovered that the display of hours for the spinner would mod 12 the hour based on the format containing AM/PM, which should be a 12 hour clock, but it seemed technically more correct to use the new 12-hour clock field instead. I also noticed that I wasn't passing a prop for the max spinner for time range, so I fixed that as well.

Issue #17854, opened 2019-10-25
[Bug Report] notification can't display all text when there are some long words in the message


xijiezhong:
### Element UI version
2.12.0

### OS/Browsers version
mac hight sierra 10.13.4

### Vue version
2.6.10

### Reproduction Link
https://jsfiddle.net/z3o04qvk/3/

### Steps to reproduce
just refresh the page, every time the notification show out will replay the bug

### What is Expected?
the notification dialog show the word completely

### What is actually happening?
the notification dialog just show a part of the long words

PR #17855, opened 2019-10-25
Notification: fix long word cannot display completely bug(#17854)

xijiezhong:
this PR is to fix the issue #17854 . 
When there are some long words in the message that the notification should show, the notification cannot show all the words, it just show part of the words.
Obviously, it is a bug. And this PR will fix it.


island205:
Sorry, we have no plan to support this. You can do this with custom style.

Issue #17856, opened 2019-10-25
[bu]Select component trembles on zoom


steb1193:
### Element UI version
2.12.0

### OS/Browsers version
Any / Google Chrome 72.0.3626.96

### Vue version
2.6.10

### Reproduction Link
https://codepen.io/steb1193/pen/QWWvYwW

### Steps to reproduce
zoom browser to 110% (for example)
select first 6 options

### What is Expected?
select work as normal

### What is actually happening?
select began to tremble

PR #17858, opened 2019-10-25
Select: Select component trembles on zoom (#17856)

steb1193:


island205:
PR without descriptions and refer relative issues  will be closed directly.

PR #17970, opened 2019-11-06
[autocomplete] explicit aria-expanded="false", aria-controls mapped to selection list id correctly, aria-activedescendant removed when no options are shown

Erbenos:
Google Lighthouse complains about the lack of explicit `aria-expanded="false"` on autocomplete container when the available selection list is not expanded.


Erbenos:
Additionally, aria-controls on `el-autocomplete` `input` was not mapped correctly to option list id. It was just "id".

Erbenos:
Also, `aria-activedescendant` removed when no options are shown. Google Lighthouse complained warns about the fact that attribute was set even when no such "descendants" were available to the user to select from.

kevrousso:
Is this approved for merge? Can anyone merge this if it is?

Issue #14002, opened 2019-01-07
[Bug Report] Table header horizontal scrollbar not syncing with body horizontal scrollbar


jrosczyk:
### Element UI version
2.4.11

### OS/Browsers version
Mac / Safari, Chrome, Firefox

### Vue version
2.5.21

### Reproduction Link
[https://jsfiddle.net/b6wsdhoc/](https://jsfiddle.net/b6wsdhoc/)

### Steps to reproduce
1. Go to this jsfiddle: https://jsfiddle.net/b6wsdhoc/
2. Scroll right in the table header using the scrollbar (not a gesture scroll)
3. You will see that the table body does not scroll with the header

### What is Expected?
That the table header horizontal scrollbar syncs with the body scrollbar and the body content moves with the header. The same way that the body scrollbar syncs and moves the header scrollbar. Both ways should be in sync.

### What is actually happening?
The header horizontal scrollbar moves the column header only, not the table body wrapper. Using a touch gesture mouse and swiping does sync both... but when using the scrollbar on header the body doesn't move with it.

sstraus:
Any chance to get it fixed?

sandesh-bhoir:
this issue still persists, is there any update on same..!

PR #18286, opened 2019-12-03
Table: Fix table header scroll bug (#14002)

daochouwangu:
Fix the table body don't scroll bug when move table header or table footer  scrollbar
#14002


daochouwangu:
会导致默认情况下table header的mousewheel失效。因为hidden和visible的table header 并不会触发scroll事件

daochouwangu:
加回Mousewheel 这样header没有scroll的时候也能滚动

add back Mousewheel event ,so header can scroll when it don't have scroll event

PR #18348, opened 2019-12-09
Message: fix Error instance signed to Ctor.data

koucxz:
When an error seted to the first param, like the Message(error).it occurs when promises uncaught error rejected, and it's resolving npm packages errors.
```
axios().catch(error => Message(error))
```
[Vue warn]: data functions should return an object:

and serveral big errors.


koucxz:
i didn't do a test, now there's sth wrong

PR #18439, opened 2019-12-17
[Input Number] Correctly compute inputNumberDisabled when false

### LINKED ISSUE #18440: [Bug Report] [Input Number] aria-disabled is incorrectly set to "undefined" causing to be truthy
### Element UI version
2.13.0

### OS/Browsers version
OSX/Chrome

### Vue version
2.6.11

### Reproduction Link
https://jsfiddle.net/9np2wcde/

### Steps to reproduce
The issue can also be seen in the docs here, https://element.eleme.io/#/en-US/component/input-number#inputnumber

Create any `number-input` component not inside of a form and do not pass a disabled prop. See `aria-disabled="undefined"`.

### What is Expected?
It should say `aria-disabled=false` or have no `aria-disabled` attribute at all.

It has an `aria-disabled` attribute that is set to `"undefined"`

ashuser-pendo:
`input-number.vue` incorrectly calculates `inputNumberDisabled` when the prop `disabled` is set to false as well as `elForm` not existing. `inputNumberDisabled` returns `"undefined"` instead of `false` which is treated as a truthy value on `aria-disabled`.

No issue has been created, I just noticed the issue while debugging some of my own code. Will create and link issue if that is expected.

Fixed #18440

### DIFF
--- a/packages/input-number/src/input-number.vue
+++ b/packages/input-number/src/input-number.vue
@@ -168,7 +168,7 @@
         return this.size || this._elFormItemSize || (this.$ELEMENT || {}).size;
       },
       inputNumberDisabled() {
-        return this.disabled || (this.elForm || {}).disabled;
+        return this.disabled || !!(this.elForm || {}).disabled;
       },
       displayValue() {
         if (this.userInput !== null) {


bcruddy:
👍 LGTM

PR #18527, opened 2019-12-25
Fix "Cannot read property 'level' of null"

MSharifi78:
When we are trying to change `value` after updating options, an exception will be thrown with the following message:
```javascript
[Vue warn]: Error in callback for watcher "value": "TypeError: Cannot read property 'level' of null"

found in

---> <ElCascaderPanel> at packages/cascader-panel/src/cascader-panel.vue
       <ElCascader> at src/components/element-ui/Cascader.vue
```


pmzi:
Same here.

We are having this error when our value is changed simultaneously with options. It seems that options is set after value (But we have changed options before value in our side).

PR #18758, opened 2020-01-27
[Step] Use steps $parent property instead of $children

Tofandel:
When using `el-steps` with elements other than `el-step` (To display content when vertical for instance)

When the step is changed, an error is thrown (`TypeError: prevChild.calcProgress is not a function`) because the element is using the `$children` property of the `$parent` (Refering to the el-step components *and also* the content elements that are not el-step and thus messing up the index of the `$children` property)

The el-step element is already adding a property `steps` to it's parent and this property should be used instead of `$children`

This fixes the previously described bug

### DIFF
--- a/packages/steps/src/step.vue
+++ b/packages/steps/src/step.vue
@@ -136,7 +136,7 @@ export default {
 
   methods: {
     updateStatus(val) {
-      const prevChild = this.$parent.$children[this.index - 1];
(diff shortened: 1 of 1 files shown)


Tofandel:
Bump

Tofandel:
@cs1707 Can you have a look please?

PR #19536, opened 2020-05-31
[Form] Adding capability to validate part of the fields while still getting a promise in return

athrunsun:

### This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [x] TypeScript definition update
- [x] Refactoring
- [ ] Code style optimization
- [ ] Branch merge
- [ ] Other (about what?)

### What's the background?
- I need to validate part of the fields and expect a `Promise` as return value.
- `Form.validate()` supports returning a promise, but doesn't support validating part of the fields. From my understanding, this method is to get a final result against all fields' validity.
- `Form.validateField()` supports validating part of the fields, but doesn't support returning a promise. From my understanding, this method is to apply a callback to each of the fields passed in.

### API Implementation
My original goal is to get a final result against **part of the fields' validity**, thus refactoring `Form.validate()` is my choice.

I added a second parameter to this method, making it `validate(callback, fieldProps)`. In order to avoid a destructive refactoring, I put `fieldProps` at the 2nd position - inserting it at the 1st position will break this method's backward compatibility, which is unacceptable IMO.

### What's the effect?
Now I can validate just part of the form fields while still getting a `Promise` in return,

```js
const valid = await this.$refs.dataForm.validate(undefined, ['fieldProp1', 'fieldProp2']);
```

### Changelog description
1. English description
> Add a second parameter `fieldProps` to `Form.validate()` to support validating part of the form fields and get a `Promise` in return.

2. Chinese description (optional)
> 给 `Form.validate()` 增加第二个参数 `fieldProps` 以支持验证部分字段的同时也能获得一个 `Promise` 作为返回值。

### Self Check before Merge

- [ ] Doc is updated/provided or not needed
- [x] Demo is not needed
- [x] TypeScript definition is updated/provided
- [x] Changelog is provided

### DIFF
--- a/packages/form/src/form.vue
+++ b/packages/form/src/form.vue
@@ -106,10 +106,19 @@
           field.clearValidate();
         });
       },
-      validate(callback) {
+      validate(callback, fieldProps) {
         if (!this.model) {
           console.warn('[Element Warn][Form]model is required for validate to work!');
           return;
+        }
+
+        fieldProps = [].concat(fieldProps);
+        let fields;
+        // 没传入fieldProps或者传入的为空数组时，验证所有字段
+        if (fieldProps.length === 0) {
+          fields = this.fields;
+        } else {
+          fields = this.fields.filter(field => fieldProps.indexOf(field.prop) !== -1);
         }
 
         let promise;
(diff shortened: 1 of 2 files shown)


athrunsun:
One question: where can I update the documentation site?

CarterLi:
Why not modify `validateField`?

athrunsun:
@CarterLi Please look carefully into "What's the background?" and "API Implementation" sections. From my understanding, although those 2 methods both receive callback as argument, they serve different purposes.

PR #19982, opened 2020-07-31
Select: fixed hoverIndex calculation error

wangmeijian:
1. Solved the problem that the Select component that supports remote search selects inconsistent options before and after changes in options data
2. When v-for traversal generates option, do not configure key. Add pr this time, and solve issue#19557


xianghongai:
Bug 照样存在

PR #20022, opened 2020-08-07
fix: scss file compile error when using dart-sass

roojay520:
当使用 `dart-sass` 对项目的主题文件进行编译时, 双引号会导致编译失败

elementui-scss-error


FrankCheungCN:
I wonder why this commit is not deployed yet?

PR #20509, opened 2020-12-01
Dialog: Added wrapperClass prop

vschst:
Added `wrapperClass` prop to customize dialog wrapper element.

### DIFF
--- a/packages/dialog/src/component.vue
+++ b/packages/dialog/src/component.vue
@@ -5,7 +5,7 @@
     @after-leave="afterLeave">
     <div
       v-show="visible"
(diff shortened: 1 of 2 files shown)


packages/dialog/src/component.vue
```diff
@@ -5,7 +5,7 @@
     @after-leave="afterLeave">
     <div
       v-show="visible"
-      class="el-dialog__wrapper"
+      :class="['el-dialog__wrapper', wrapperClass]"
```
shamilfrontend, line 8: `computed` ?)

PR #20626, opened 2020-12-29
Drawer: fix focus bug

cs1707:

ref: #19902

### DIFF
--- a/packages/drawer/src/main.vue
+++ b/packages/drawer/src/main.vue
@@ -26,7 +26,7 @@
           >
           <header class="el-drawer__header" id="el-drawer__title" v-if="withHeader">
             <slot name="title">
-              <span role="heading" tabindex="0" :title="title">{{ title }}</span>
+              <span role="heading" :title="title">{{ title }}</span>
             </slot>
             <button
               :aria-label="`close ${title || 'drawer'}`"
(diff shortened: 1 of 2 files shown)


JeremyWuuuuu:
Hey, I am the contributor who contributed this component, the original idea of adding `tabindex` to `title` is because we want the screenreader to read the title content for the users who uses screenreader device, but I did have made a mistake on adding `tabindex` towards the container.

JeremyWuuuuu:
The change itself was legit though, but still, I think you do not need to remove the `nextTick` operation because that's how it works, also I didn't add `trap-focus` for this component, which needs trap the focus(keyboard navigation) inside this content.

cs1707:
Hi, Thanks for your reply. In my opinion that `tabindex` can improve keyboard accessibility. For screen reading we should use `aria-label` or `role` attributes to improve accessibility. So I removed the `tabindex` from the title.

cs1707:
For dialog or drawer components I think trap-focus is necessary. We have the same opinion about this. And like the bootstrap antd material ui they all have implementations for it.

JeremyWuuuuu:
> For dialog or drawer components I think trap-focus is necessary. We have the same opinion about this. And like the bootstrap antd material ui they all have implementations for it.

The change is legit, adding `trap-focus` functionality would be a great upgrade.

cs1707:
> > For dialog or drawer components I think trap-focus is necessary. We have the same opinion about this. And like the bootstrap antd material ui they all have implementations for it.
> 
> The change is legit, adding `trap-focus` functionality would be a great upgrade.

This feature may be updated next time 😄

Issue #19370, opened 2020-05-09
[Bug Report] Memory leak at el-tooltip cleanup


badpunman:
### Element UI version
0.1.0

### OS/Browsers version
All

### Vue version
3.0.0-beta.10

### Reproduction Link
https://github.com/ElemeFE/element/blob/dev/packages/tooltip/src/main.js

### Steps to reproduce
Traced a memory leak to this to the code:
https://github.com/ElemeFE/element/blob/dev/packages/tooltip/src/main.js

in the mounted there is a 'focus' event listener registered to lambda. In the destroyed we try to to clean it by referencing 'this.handleFocus' but that is not the function that is bound to the event. 

### What is Expected?
The component event listener should be cleaned up properly.

### What is actually happening?
Event is not cleaned up and we leak memory.

Michael-Zhang-Xian-Sen:
have same question

lianzhao:
Not popover only...
![image](https://user-images.githubusercontent.com/536786/103114460-9a717700-4699-11eb-90a6-06b5e7513378.png)

PR #20671, opened 2021-01-11
Fix memory leak issues in tooltip & popover

lianzhao:

Fix memory leak issues in `tooltip` & `popover`

Fix https://github.com/ElemeFE/element/issues/19370
Fix https://github.com/ElemeFE/element/issues/18148

### DIFF
--- a/packages/tooltip/src/main.js
+++ b/packages/tooltip/src/main.js
@@ -113,18 +113,7 @@ export default {
       this.$el.setAttribute('tabindex', this.tabindex);
       on(this.referenceElm, 'mouseenter', this.show);
       on(this.referenceElm, 'mouseleave', this.hide);
-      on(this.referenceElm, 'focus', () => {
-        if (!this.$slots.default || !this.$slots.default.length) {
-          this.handleFocus();
-          return;
-        }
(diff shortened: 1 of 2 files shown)


packages/tooltip/src/main.js
```diff
@@ -157,6 +146,18 @@ export default {
...
     },
     handleFocus() {
+      if (!this.$slots.default || !this.$slots.default.length) {
+        this.doFocus();
+        return;
+      }
+      const instance = this.$slots.default[0].componentInstance;
+      if (instance && instance.focus) {
+        instance.focus();
+      } else {
```
carterbs, line 156: I think we should only 'doFocus' if instance exists. I was considering putting in a PR for this exact issue, which is causing tooltips to show when certain elements re-render in my app. What do you think @lianzhao?
```suggestion
      } else if (instance) {
```

xgenvn: @lianzhao can you take a look at this? We are experiencing the same issues.

lykiao: this issue haven't be resolved? too many detached! who can tell me will publish to which version?
image

Issue #14936, opened 2019-04-02
[Feature Request] Trigger collapse by clicking on arrow icon/button


japhfortin:
### Existing Component
Yes

### Component Name
collapse

### Description
Hi,
I think it would be good if we can have some option to allow the user to only trigger the collapse by clicking the arrow icon/button instead of the entire header.

iamkun:
Sorry, we have no plan to support this.

luck2011:
Yes, users need this feature when they have title slots on Collapse!
They donot want the title contents trigger the collapse event.

PR #20903, opened 2021-04-02
Collapse: add icon trigger attribute

luck2011:
Collapse adds the `trigger` attribute.  
When users use custom content with title `slot`, they donot want the title triggers the collapse event, but only the icon arrow. See #14936 
Now setting `trigger="icon"` on the `<el-collapse />` or `<el-collapse-item />` components will help.


pyaehein:
+1

PR #21103, opened 2021-06-25
the table $emit('selection-change') should not return data of dumplicate rowKey

Valar103769:
The incoming data should be deduplicated using id instead of judging the reference address


Valar103769:
For compatible  prop `reserve-selection`

PR #21105, opened 2021-06-25
Table: fix reserve-selection bug

Valar103769:
When reserve `selection = false`,· @selection-change`  throws a value of checked to true. However, when `reserve selection = true`, `@selection-change` throws a value of checked to true and false. Consider a scenario in which the dialog sets a table and refreshes tabledata every time `dialogvisible = true`. In this case, the `indexof (row)` in the source code always returns false, It will lead to state error, and it should be judged according to rowkey

### DIFF
--- a/packages/table/src/store/watcher.js
+++ b/packages/table/src/store/watcher.js
@@ -156,7 +156,9 @@ export default Vue.extend({
     },
 
     toggleRowSelection(row, selected, emitChange = true) {
-      const changed = toggleRowStatus(this.states.selection, row, selected);
+      const rowKey = this.states.rowKey;
+      const rowId = getRowIdentity(row, rowKey);
+      const changed = toggleRowStatus(this.states.selection, row, selected, rowId);
       if (changed) {
         const newSelection = (this.states.selection || []).slice();
         // 调用 API 修改选中值，不触发 select 事件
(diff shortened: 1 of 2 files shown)


cs1707:
> When reserve `selection = false`,· @selection-change` throws a value of checked to true. However, when`reserve selection = true`, `@selection-change`throws a value of checked to true and false. Consider a scenario in which the dialog sets a table and refreshes tabledata every time`dialogvisible = true`. In this case, the `indexof (row)` in the source code always returns false, It will lead to state error, and it should be judged according to rowkey

hi, @Valar103769 , Thank you for submitting the pr. Would you like to use jsfiddle/codepen/codesandbox to reproduce the issue?

Valar103769:
> > When reserve `selection = false`,· @selection-change` throws a value of checked to true. However, when`reserve selection = true`, `@selection-change`throws a value of checked to true and false. Consider a scenario in which the dialog sets a table and refreshes tabledata every time`dialogvisible = true`. In this case, the `indexof (row)` in the source code always returns false, It will lead to state error, and it should be judged according to rowkey
> 
> hi, @Valar103769 , Thank you for submitting the pr. Would you like to use jsfiddle/codepen/codesandbox to reproduce the issue?

[codepen](https://codepen.io/valar123/pen/KKmJXgR)

ps： 例子中用深拷贝，模拟服务器请求的数据

步骤： 选中几项， 然后点击 从服务器拉取数据来更新 按钮，模拟数据请求， 会发现选中状态丢失

业务场景：当配合翻页组件使用时， 每次翻页都会从服务器重新拉取数据，希望能保留住之前选中的项，
期望：  翻到之前有选中项的页码时， 希望设置相应的选中项
实际效果：选中项丢失了

PR #21158, opened 2021-07-17
Form: Fix bug in validate function 

zeyadmoamen1210:
in form-item file in validate function I trimed the input field value from spaces because that make required rule not accurate 
it can be ignored with  only spaces

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,17661 @@
+{
+  "name": "element-ui",
+  "version": "2.15.3",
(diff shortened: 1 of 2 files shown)


cs1707:
I don't think this is a good idea. Ignoring spaces is not the behavior most users expect

zeyadmoamen1210:
that's not make sense to skip required validate with only spaces and  save only spaces in database !

PR #21224, opened 2021-08-12
fix: theme-chalk mixin compilation error with dart-sass

haywirez:
Should fix compilation error when running `sass` instead of `node-sass`:

```Module build failed (from ./node_modules/sass-loader/dist/cjs.js):

    @media only screen and #{inspect(map-get($map, $key))} {
                          ^
      Expected "(".
  ╷
9 │     @media only screen and "(min-width: 768px) and (max-width: 991px)"{
  │                            ^
  ╵
  node_modules/element-ui/packages/theme-chalk/src/mixins/mixins.scss 9:28  res()
```


zakariaf:
@haywirez as they didn't merge this PR yet, and I have this error too; is there a different solution?

PR #21260, opened 2021-08-24
fix(upload): 修复进度条引起跨域导致文件上传失败问题

gxdvip:
由于WEB后端配置CORS未处理OPTION，upload进度条效果引起跨域导致文件上传失败，故增加隐藏进度条选项来解决：回调函数onProgress() {return false} 即可

### DIFF
--- a/packages/upload/src/ajax.js
+++ b/packages/upload/src/ajax.js
@@ -36,7 +36,7 @@ export default function upload(option) {
   const xhr = new XMLHttpRequest();
   const action = option.action;
 
-  if (xhr.upload) {
+  if (xhr.upload && option.onProgress(false) !== false) {
     xhr.upload.onprogress = function progress(e) {
       if (e.total > 0) {
         e.percent = e.loaded / e.total * 100;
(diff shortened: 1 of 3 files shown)


cs1707:
A bit confused by this solution. I think the correct solution would be to have the backend support `OPTIONS` method.
> In CORS, a preflight request is sent with the OPTIONS method so that the server can respond if it is acceptable to send the request. In this example, we will request permission for these parameters:
>
> The Access-Control-Request-Method header sent in the preflight request tells the server that when the actual request is sent, it will have a POST request method.
The Access-Control-Request-Headers header tells the server that when the actual request is sent, it will have the X-PINGOTHER and Content-Type headers. 

ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS

gxdvip:
@cs1707 确实应该由后端解决OPTION，但不是必须的，因为preflight可以跳过。
最初用upload组件 上传报错 ：
`
Access to XMLHttpRequest at 'https://dev.ex.com/api/uploadFile' from origin 'http://localhost:63342' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
`
而用axios可以上传成功。
网上其他解决方案是在before-upload 或 http-request 改用axios。

这个解决方案是简单的在on-progress方法里 return false 即可像axios一样上传成功。

PR #21951, opened 2022-06-28
Update picker.vue

jacky-kang:
vue.runtime.esm.js:619 [Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "placement"


whoisarpit:
Facing the same issue, thanks for the fix

Issue #22175, opened 2022-09-29
[Bug Report] Parsing column prop with dots or barckets 


i33558493:
### Element UI version
2.15.10

### OS/Browsers version
Windows 10 / Chrome 105.0.5195.127

### Vue version
2.7.10

### Reproduction Link
https://codepen.io/i33558493/pen/PoeQgdw

### Steps to reproduce
1. Create a el-table component.
2. Set the column prop like 'a.b' or 'a[b]'.
3. Make some table data wtih a property name like 'a.b' or 'a[b]'.

### What is Expected?
Show the 'a.b' column.

### What is actually happening?
Data of 'a.b' column do not show.

i33558493:
I noticed the discussion in this issue #5101. Now I know there is a feature to parse data path.But I think the table component can parse prop smartly, just like [element plus](https://github.com/element-plus/element-plus).
I expecte the table component will show the 'a.b' column if the data object has the 'a.b' prooerty.Also, it will parsing the path if the data object doesn't have the 'a.b' prooerty.

For this, we can change the function `getPropByPath` in `src\utils\util.js`.
```javascript
export function getPropByPath(obj, path, strict) {
+  if (path in obj) {
+    return {
+      o: obj,
+      k: path,
+      v: obj ? obj[path] : null
+    };
+  }

  let tempObj = obj;
  path = path.replace(/\[(\w+)\]/g, '.$1');
  path = path.replace(/^\./, '');

  let keyArr = path.split('.');
  let i = 0;
  for (let len = keyArr.length; i < len - 1; ++i) {
    if (!tempObj && !strict) break;
    let key = keyArr[i];
    if (key in tempObj) {
      tempObj = tempObj[key];
    } else {
      if (strict) {
        throw new Error('please transfer a valid prop path to form item!');
      }
      break;
    }
  }
  return {
    o: tempObj,
    k: keyArr[i],
    v: tempObj ? tempObj[keyArr[i]] : null
  };
};
```

dgendill:
[i33558493](https://github.com/i33558493) I also think it would be great if props could support brace notation. I'm would also like to have support for object keys with special characters. For me, I have data with object keys that contains periods and semicolons, for example:

```
{
  mountains: {
    'Mt. Fiji': 1,
    'Mountain: Mt. Blanca': 2
  }
}
```

Here is my idea to update `util.getPropByPath` with a non-breaking change to support braces and object keys with special characters. I think this way provides backward compatibility for the current implementation also.

[...]

Issue #17893, opened 2019-10-29
[Bug Report] 对象键值包含 . 时 el-form-item 配置 prop 报错：Error: please transfer a valid prop path to form item!


chenquincy:
### Element UI version
2.12.0

### OS/Browsers version
MacOS，Chrome最新

### Vue version
2.6.10

### Reproduction Link
https://jsfiddle.net/mmx38qxw/

### Steps to reproduce
1. 使用 el-form 表单，绑定对象 { 'test.test': 'test' }
2. el-form-item 设置 prop 为 'test.test'

### What is Expected?
正常使用，prop通过 xx.xx 检验嵌套属性是个不错的功能，但是看了源码发现貌似没办法满足键值实际上就是 xx.xx 的情况

### What is actually happening?
报错 Error: please transfer a valid prop path to form item!

my9074:
+1

@chenquincy 目前有解决方案吗？

sid3r:
Element uses [yiminghe/async-validator](https://github.com/yiminghe/async-validator) to validate form data, give it a try.

jinweizhiyuan:
I have the same problem

jinweizhiyuan:
have a  the solution to this problem ?

Issue #10293, opened 2018-03-21
[Feature Question]Why does component el-form-item's prop use format like 'arr.index.key', rather than 'arr[index].key' in documents?


sky111144:
#### [Question] Why does component el-form-item's property prop use format like 'arr.index.key', rather than 'arr[index].key' in documents of Element-UI v2.20?
+ I am  curious about that component el-form-item's  prop use format like 'arr.index.key'. Because I have read source code of function getPropByPath that is how to handle el-form-item's property  prop. Function getPropByPath also can handle format like 'arr[index].key'.
+ This question happened when I read documents of Element-UI' el-form-item.
+ Documents
``` html
// demo in http://element-cn.eleme.io/#/zh-CN/component/form
  <el-form-item
    v-for="(domain, index) in dynamicValidateForm.domains"
    :label="'域名' + index"
    :key="domain.key"
    :prop="'domains.' + index + '.value'"
    :rules="{
      required: true, message: '域名不能为空', trigger: 'blur'
    }"
  >
    <el-input v-model="domain.value"></el-input><el-button @click.prevent="removeDomain(domain)">删除</el-button>
  </el-form-item>
```
+ Source code:
``` JavaScript
// line 218 in  https://github.com/ElemeFE/element/blob/dev/packages/form/src/form-item.vue
let prop = getPropByPath(model, path, true);
```
``` JavaScript
//  line 44 - 68 in https://github.com/ElemeFE/element/blob/dev/src/utils/util.js
export function getPropByPath(obj, path, strict) {
  let tempObj = obj;
  path = path.replace(/\[(\w+)\]/g, '.$1');
  path = path.replace(/^\./, '');

  let keyArr = path.split('.');
  let i = 0;
  for (let len = keyArr.length; i < len - 1; ++i) {
    if (!tempObj && !strict) break;
    let key = keyArr[i];
    if (key in tempObj) {
      tempObj = tempObj[key];
    } else {
      if (strict) {
        throw new Error('please transfer a valid prop path to form item!');
      }
      break;
    }
  }
  return {
    o: tempObj,
    k: keyArr[i],
    v: tempObj ? tempObj[keyArr[i]] : null
  };
};
```

PR #22301, opened 2022-12-04
Update getPropByPath to support braces and object keys with special characters #22175, #10293, and #17893

dgendill:

This PR is related to #22175, #10293, and #17893. Although it's rare, some of us need to lookup props using brace notation and keys that contain special characters. For example when our data looks like this and we need to validate the mountains['Mountain: Mt. Blanca'] property.

```
{
  mountains: {
    'Mt. Fiji': 1,
    'Mountain: Mt. Blanca': 2
  }
}
```

This PR updates getPropByPath so it can parse object key paths using brace notation and parse keys containing special characters. It also adds test cases for the getPropByPath function.

I do have an outstanding question related to this code in form-item.vue.

```
if (path.indexOf(':') !== -1) {
  path = path.replace(/:/, '.');
}
```

This code will prevent the use of semicolons in object keys. Do you think there is a way to safely remove this? Is there an alternative? This conversions seem like an undocumented feature that the users shouldn't be relying on, but I don't know the code well enough to be certain.

### DIFF
--- a/packages/form/src/form-item.vue
+++ b/packages/form/src/form-item.vue
@@ -149,10 +149,6 @@
         if (!model || !this.prop) { return; }
 
         let path = this.prop;
-        if (path.indexOf(':') !== -1) {
-          path = path.replace(/:/, '.');
-        }
-
         return getPropByPath(model, path, true).v;
       },
       isRequired() {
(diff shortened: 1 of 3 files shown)


k644606347:
Why hasn't there been any progress

PR #22451, opened 2023-03-17
input: icon toggle between show and hide password

Dunky-Z:
Icon library does not have closed-eye icons,
use el-icon-minus to simulate closed-eye icons.

Closes: #22316

### DIFF
--- a/packages/input/src/input.vue
+++ b/packages/input/src/input.vue
@@ -65,7 +65,8 @@
             @click="clear"
           ></i>
           <i v-if="showPwdVisible"
-            class="el-input__icon el-icon-view el-input__clear"
+            class="el-input__icon el-input__clear"
+            :class="passwordVisible ? 'el-icon-view'  : 'el-icon-minus'"
             @click="handlePasswordVisible"
           ></i>
           <span v-if="isWordLimitVisible" class="el-input__count">


Dunky-Z:
Demo: https://codepen.io/Dunky-Z/pen/JjaZZxV

webvs2:
I will go to the ui group to apply for an appropriate icon. I suggest that we keep the changes and update them later

Dunky-Z:
Hey webvs2, thanks for considering my PR! I totally understand your point about trying to get a suitable icon from the UI group, but I think it's worth keeping my changes in the meantime. Applying for a new icon from the Unicode organization, which could take a considerable amount of time. The password hide/show icon is a very common feature and it's important that users can easily tell whether the password is being hidden or shown. If we don't make this change, users might assume there's a bug with the code, as was mentioned in Issue 22316.

As for the visual style of the icon, I think the minus sign I added is simple but effective, and it's easy for developers to modify if they want something different. Plus, if a better icon becomes available in the future, we can always update it as an optimization later on.

So what do you say? Could we go ahead and accept my PR with the current changes? Thanks again for your help!

webvs2:
Yes, that's a good thought. There's still time until the next release. Maybe the new icon will catch up.

PR #22566, opened 2023-06-13
Table: fix the background style is incomplete when merging row hover …

y1huachen:
…(#22514)

Referring to the hover implementation of merging rows in tables in antd, dynamically add styles when hovering over a cell with the mouse.


wangdaodao:
`handleCellMouseEnter`和`handleCellMouseLeave`加一下防抖是不是会好点？

PR #22968, opened 2024-09-13
fix: Repair the alignment error problem when label-width is auto and …

wxwzl:
fix: Repair the alignment error problem when label-width is auto and label-position is left


packages/form/src/label-wrap.vue
```diff
@@ -17,7 +17,11 @@ export default {
       if (autoLabelWidth && autoLabelWidth !== 'auto') {
         const marginLeft = parseInt(autoLabelWidth, 10) - this.computedWidth;
         if (marginLeft) {
-          style.marginLeft = marginLeft + 'px';
+          if(this.elForm.labelPosition === 'left'){
```
wangdaodao: 使用三元运算会简洁一些
`style[(this.elForm.labelPosition === 'left') ? 'marginRight' : 'marginLeft'] = marginLeft + 'px';`

wxwzl: 嗯嗯，改了

